PR #29, opened 2019-12-15
Tweaking limits on doors so they can close

mxgrey:
The simulated sliding doors were having trouble closing because the `0.01` limit was making them crash into each other.


mxgrey:
Let's hold off on merging this, because I'm noticing some other (probably related) issues that I don't have an explanation for yet.

mxgrey:
I've added some inertial properties which should be more realistic than the defaults. This wasn't the cause of the issue I was seeing, but it should still improve the fidelity of the simulation.

PR #47, opened 2020-02-03
Bug/fix dict illegal accesses

aaronchongth:
* added some `None` initializations

* added some checks before accessing

These should fix problems for super minimal maps that have no doors, or some elements


aaronchongth:
oof, yeah I didn't dig through it deep enough to figure out if they were basic maps/dictionaries or some other fancy `collection`. Sure thing.

PR #86, opened 2020-03-12
check if key in dict first

aaronchongth:
* this will prevent any issues when opening older `.building.yaml` files, that do not have the newer key names


aaronchongth:
I would have thought it would resolve to `None` as well, maybe Im mixing up the `yaml-cpp` syntax. thanks!

PR #100, opened 2020-03-20
branch on double swing door direction for sim generation

codebot:
* branch depending on sign of `motion_direction` for double hinged doors when generating simulation model
 * use the `motion_degrees` parameter when generating double hinged doors
 * add a test map package that has a pair of double swing doors (one with positive motion direction, one with negative motion direction) that will crash into vending machines if they open the incorrect direction
 * fix an unrelated but annoying bug when trying to move models in the GUI when explicit scale or measurements were not provided


aaronchongth:
just checked with varying door types in `office` world, it works great! Thanks for the quick fix!

PR #102, opened 2020-03-23
added fixed joint offset for swing doors

aaronchongth:
* added fixed offset for swing door joints, shifting them slightly inwards to prevent visually clipping through walls when opening
* current fixed offset is 3 times the door's width

* one side-effect, the opening of the door will shrink slightly as a result, existing narrow doorway cases might have robots unable to pass through (so far none in `rmf_demos` have this edge case)
* another side-effect, the work space of the doors will be slightly smaller from what is drawn by `traffic_editor`, which should be fine

https://github.com/osrf/traffic_editor/issues/101


aaronchongth:
Closing this due to it being a nice-to-have, and we should leave it to the users to determine the level of detail they want.

PR #104, opened 2020-03-25
Feature/thumbnail generation

aaronchongth:
* added and documented the steps and pipeline of generating thumbnails to be used in `traffic_editor`, migrated from the archived 'rmf_editor`

* solves https://github.com/osrf/traffic_editor/issues/65


aaronchongth:
always fun to learn something new about the tools we use! thanks for starting this initiative!

PR #107, opened 2020-03-29
Ign plugins formatted

Yadunund:
Linted using RMF style guide to serve as a reference.


building_sim_plugins/building_sim_ign/src/door.cpp
```diff
@@ -332,12 +341,14 @@ class IGNITION_GAZEBO_VISIBLE DoorPlugin
...
+IGNITION_ADD_PLUGIN(
+  DoorPlugin,
+  System,
+  DoorPlugin::ISystemConfigure,
+  DoorPlugin::ISystemPreUpdate)
 
 // TODO would prefer namespaced
-IGNITION_ADD_PLUGIN_ALIAS(DoorPlugin,
-                          "door")
+IGNITION_ADD_PLUGIN_ALIAS(
```
luca-della-vedova: I wonder why in this case it was not changed to just be a single line since it would have fit the character count?

building_sim_plugins/building_sim_ign/src/slotcar.cpp
```diff
@@ -220,10 +229,12 @@ void SlotcarPlugin::PreUpdate(const UpdateInfo& info, EntityComponentManager& ec
...
+IGNITION_ADD_PLUGIN(
+  SlotcarPlugin,
+  System,
+  SlotcarPlugin::ISystemConfigure,
+  SlotcarPlugin::ISystemPreUpdate)
 
-IGNITION_ADD_PLUGIN_ALIAS(SlotcarPlugin,
-                          "slotcar")
+IGNITION_ADD_PLUGIN_ALIAS(
+  SlotcarPlugin,
```
luca-della-vedova: Same as above, could probably fit in one line

building_sim_plugins/building_sim_ign/src/door.cpp
Yadunund: I do not expect the linter to perform any line width optimizations on its own. It only corrects for indentations

luca-della-vedova: Hum can you put it into one line? Then we can merge this

PR #109, opened 2020-04-01
Ign plugin door

Yadunund:
Modularized the gazebo and ign door plugins while following @luca-della-vedova's pattern with slotcar.

To be merged after #107.


luca-della-vedova:
Can we rebase this to after the linting PR was merged? Most of the diff now is due to linting and it is making it a bit hard to review.

Also it doesn't compile:
```
/home/luca/demos_ws/src/rmf/traffic_editor/building_sim_plugins/building_sim_ign/src/door.cpp:201:31: error: ‘params’ was not declared in this scope
         left_door_joint, ecm, params);
                               ^~~~~~
/home/luca/demos_ws/src/rmf/traffic_editor/building_sim_plugins/building_sim_ign/src/door.cpp:217:32: error: ‘params’ was not declared in this scope
         right_door_joint, ecm, params, true);
```

I think that's because you removed the params declaration in that function to move it to the common class.

General feedback is that I feel there is much more that can be pushed to the common class, I was hoping it would turn out even simpler than slotcar, where the only functions not present in the common class are the ones that are ignition / gazebo specific.
Examples of parts that can be moved:

- State publication logic
- Most of the "common" door functions, for example open, close, calculating movements. What I'm thinking is you can have each door keep track of its position, at every Update call update joint positions through gazebo / ignition specific functions, then all the rest can be done in the common class (check if closed, calculate how much it should move).

This should reduce a lot the size of the plugins code.

Yadunund:
Thanks for the feedback @luca-della-vedova. I've pushed the publishing logic as well as the state calculations to the door_common file. Let me know what you think.

Yadunund:
> High level comment is that I'm not a big fan of the idea of using shared pointers all over the code to take care of the fact that we don't know how many doors there are at runtime (it could be one or two).
> Imho I would use a structure that allows iteration and then we wouldn't need the pointer magic to check whether doors exist or not?
> 
> Previously it was an array but I agree with the refactoring that it doesn't really help readability (is 0 right or left? is 1 right or left?).
> An improvement could probably be using hash maps, i.e. unordered_map<std::string, DoorElement>, requests could be an unordered_map<std::string, DoorUpdateRequest> etc. then you don't need to constantly check for null pointers but you can just do a range based iteration on the data structures?
> 
> Does this make sense?

Yes that makes sense. I initially began with implementing hash maps as described but found it equally verbose partly because the `door.cpp` is confirming whether the joint exists in the simulation mode, before "adding" them into `doors_common`. Hence there would still be some element of checking if the door exists. I think a good solution could be to implement hash maps and move [all this](https://github.com/osrf/traffic_editor/blob/adf4ceb19576f04d77fc181e8c692bccbc19f382/building_sim_plugins/building_sim_gazebo/src/door.cpp#L55) into `make()`. We can infer the joint limits from the sdf as you suggested. Then after calling `make()` in `door.cpp`, we check if the joints exist in the simulation via assert statements.

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
```diff
@@ -0,0 +1,225 @@
...
+  std::string right_door_joint_name() const;
+
+  void publish_state(const uint32_t door_value, const rclcpp::Time time);
+
+  bool is_initialized() const;
+
+  MotionParams& params();
+
+  void add_left_door(
+    const bool debuggable,
```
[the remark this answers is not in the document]
Yadunund: Done

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
```diff
@@ -0,0 +1,225 @@
...
+  void add_right_door(
+    const bool debuggable,
+    const double upper_limit,
+    const double lower_limit);
+
+  bool all_doors_open();
+
+  bool all_doors_closed();
+
+  double get_door_velocity(
```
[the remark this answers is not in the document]
Yadunund: Renamed to `calculate_target_velocity`

building_sim_plugins/building_sim_common/src/door_common.cpp
```diff
@@ -0,0 +1,271 @@
...
+{
+  std::unique_lock<std::mutex> lock(_mutex);
+  _state.door_name = door_name;
+  return _state;
+}
+
+std::string DoorCommon::door_name() const
+{
+  return _state.door_name;
+}
```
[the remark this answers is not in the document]
Yadunund: You're right these function are no longer being used after introducing the `make()` function. They have been removed.

building_sim_plugins/building_sim_common/src/door_common.cpp
```diff
@@ -0,0 +1,271 @@
...
+{
+  return _initialized;
+}
+void DoorCommon::publish_state(const uint32_t door_value,
+  const rclcpp::Time time)
+{
+  if (!_initialized)
+    return;
+
+  std::unique_lock<std::mutex> lock(_mutex);
```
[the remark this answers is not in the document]
Yadunund: Removed mutexes.

building_sim_plugins/building_sim_gazebo/src/door.cpp
```diff
@@ -5,217 +5,88 @@
...
         RCLCPP_ERROR(
           _ros_node->get_logger(),
           " -- Model is missing the left door joint [%s]",
           left_door_joint_name.c_str());
         return;
       }
-      _doors.emplace_back(_model->GetName() == "chart_lift_door",
-        left_door_joint, params);
+      _door_common->add_left_door(_model->GetName() == "chart_lift_door",
+        _left_door_joint->UpperLimit(0), _left_door_joint->LowerLimit(0));
```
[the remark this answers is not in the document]
Yadunund: The joint limits are now inferred from the parent sdf in `make()`. However with the ign version, `sdf::ElementPtr::GetParent()` method always returned a `nullptr`.I think this might be a bug. As a workaround, I'm using default values of the limits when this method fails. 
https://github.com/osrf/traffic_editor/blob/7a7984bd422248222fb26c69181da8b872ca4906/building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp#L202

Yadunund:
Thanks for the great feedback @luca-della-vedova . I take back what I said about the verbosity with iterable lists- I probably went about implementing it incorrectly. The revised files have a greater reduction in the amount of duplicate code between the `door.cpp` files for gazebo and ign. The `make()` function now parses the plugin sdf to obtain the limits of each joint. While this works for the sdf element supplied by the gazebo sim, it fails in ign. The `sdf::ElementPtr::GetParent()` method returns a nullptr. I think this is a bug with ign sdformat? Overall the plugins work as expected. Let me know what you think.

building_sim_plugins/building_sim_common/src/door_common.cpp
```diff
@@ -0,0 +1,199 @@
...
+  const std::string& left_door_joint_name,
+  const std::string& right_door_joint_name,
+  const std::array<double, 2>& left_joint_limits,
+  const std::array<double, 2>& right_joint_limits)
+: _ros_node(std::move(node)),
+  _params(params)
+{
+  if (!left_door_joint_name.empty() && left_door_joint_name != "empty_joint")
+    _doors.insert(std::make_pair(left_door_joint_name,
+      std::make_shared<DoorCommon::DoorElement>(
```
luca-della-vedova, line 63: Another "why pointer" feedback, can it just be a DoorElement? I also personally like to create a pair with the curly braces (i.e. `_doors.insert({name, element})`) since it's less verbose but that's just a preference

building_sim_plugins/building_sim_ign/src/door.cpp
```diff
@@ -180,91 +72,38 @@ class IGNITION_GAZEBO_VISIBLE DoorPlugin
     ignwarn << "Initializing plugin with name " << plugin_name << std::endl;
     _ros_node = std::make_shared<rclcpp::Node>(plugin_name);
 
-    RCLCPP_INFO(
-      _ros_node->get_logger(),
+    RCLCPP_INFO(_ros_node->get_logger(),
       "Loading DoorPlugin for [%s]",
       _model.Name(ecm).c_str());
+
     auto sdfClone = sdf->Clone();
```
luca-della-vedova, line 79: Bit of a shot in the dark here, but can you try to remove the Clone function and just use the original SDF pointer? I had a look at the sdformat source code and when cloning the parent is not copied, hence is set to default of nullptr.
We would need to make sure we are not editing the sdf element then, hopefully the utilities functions take care of that.

building_sim_plugins/building_sim_ign/src/door.cpp
```diff
@@ -180,91 +72,38 @@ class IGNITION_GAZEBO_VISIBLE DoorPlugin
...
-        left_door_joint, ecm, params);
-    }
 
-    if (right_door_joint_name != "empty_joint")
+    for (const auto& joint_name : _door_common->joint_names())
     {
-      const auto right_door_joint = _model.JointByName(ecm,
-          right_door_joint_name);
-      if (!right_door_joint)
+      const auto joint = std::make_shared<Entity>(
```
luca-della-vedova, line 91: Doesn't need to be a pointer

building_sim_plugins/building_sim_ign/src/door.cpp
```diff
@@ -180,91 +72,38 @@ class IGNITION_GAZEBO_VISIBLE DoorPlugin
...
-          right_door_joint_name.c_str());
+        RCLCPP_ERROR(_ros_node->get_logger(),
+          " -- Model is missing the joint [%s]",
+          joint_name.c_str());
         return;
       }
-      _doors.emplace_back(_model.Name(ecm) == "chart_lift_door",
-        right_door_joint, ecm, params, true);
+      create_entity_components(*joint, ecm);
+      _joints.insert(std::make_pair(joint_name, joint));
```
luca-della-vedova, line 101: Initializer list instead of make_pair

building_sim_plugins/building_sim_ign/src/door.cpp
```diff
@@ -140,26 +31,27 @@ class IGNITION_GAZEBO_VISIBLE DoorPlugin
 private:
   rclcpp::Node::SharedPtr _ros_node;
   Model _model;
+  std::unordered_map<std::string, std::shared_ptr<Entity>> _joints;
```
luca-della-vedova, line 34: At the end of the day Entity is just a uint64_t so I think it's fine to copy it around instead of making it a pointer

building_sim_plugins/building_sim_ign/src/door.cpp
```diff
@@ -140,26 +31,27 @@ class IGNITION_GAZEBO_VISIBLE DoorPlugin
 private:
   rclcpp::Node::SharedPtr _ros_node;
   Model _model;
```
luca-della-vedova, line 33: It seems now this doesn't need to be a class variable anymore since it's only used in one function?

building_sim_plugins/building_sim_ign/src/door.cpp
Yadunund: It doesn't have to be a pointer, but I'm using one so that it is analogous to the JointPtr in `door.cpp` in building_gazebo_plugin. I can change it to `Entity` if you prefer.

building_sim_plugins/building_sim_ign/src/door.cpp
Yadunund: You hit the bullseye. I now clone the sdf element ptr inside the `make()` function and work with the clone to retrieve necessary plugin parameters. The parent is then obtained from the original inside the same function. 
https://github.com/osrf/traffic_editor/blob/c518197429338d65cc5303f89819252f2da128e2/building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp#L131

building_sim_plugins/building_sim_common/src/door_common.cpp
Yadunund: Sure, I've corrected it.

building_sim_plugins/building_sim_ign/src/door.cpp
Yadunund: Not a pointer anymore.

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+    else if (joint_name == right_door_joint_name)
+    {
+      get_sdf_param_if_available<double>(element, "lower", right_joint_lower_limit);
+      get_sdf_param_if_available<double>(element, "upper", right_joint_upper_limit);
+      RCLCPP_INFO(node->get_logger(),
+        "Joint [%s] lower [%f] upper [%f]",
+        joint_name.c_str(),left_joint_lower_limit, left_joint_upper_limit);
+    }
+  };
```
luca-della-vedova, line 201: Would there be a way to remove the duplication here and make the lambda a function that returns the limits instead of modifying global variables?
I would personally remove the print statement as well but no strong feelings there.

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
```diff
@@ -0,0 +1,244 @@
...
+    right_joint_lower_limit, right_joint_upper_limit};
+
+  std::shared_ptr<DoorCommon> door_common(new DoorCommon(
+      door_name,
+      node,
+      params,
+      left_door_joint_name,
+      right_door_joint_name,
+      left_joint_limits,
+      right_joint_limits));
```
luca-della-vedova, line 238: You could simplify the DoorCommon constructor if it took a unordered_map<std::string, std::array<double,2>> that maps joint_name -> limits then you only need one argument instead of four. This map can be populated with the outputs of the refactored lambda function that returns pairs.
Also you won't need to check again in the door constructor for the validity of the joint, you can just iterate through the data structure.

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
```diff
@@ -0,0 +1,244 @@
...
+  extract_limits(joint_element);
+  // Find next joint element if present
+  joint_element = joint_element->GetNextElement("joint");
+  if (joint_element)
+    extract_limits(joint_element);
+
+  std::array<double,2> left_joint_limits ={
+    left_joint_lower_limit, left_joint_upper_limit};
+  std::array<double,2> right_joint_limits ={
+    right_joint_lower_limit, right_joint_upper_limit};
```
luca-della-vedova, line 229: Can be removed / moved inside the lambda after refactoring

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
Yadunund: Modified such that DoorCommon constructor now accepts an `std::unordered_map<std::string, DoorCommon::DoorElement>`. This is directly used to initialize `_doors`.

building_sim_plugins/building_sim_common/include/building_sim_common/door_common.hpp
Yadunund: I've removed the duplication. The `make()` function is responsible for passing an `std::unordered_map<std::string, DoorCommon::DoorElement>` object to the constructor. The lambda now helps to directly populate this object for valid joint names. I think having the lambda modify the global object is more efficient than returning a pair and inserting it later. It also handles cases where the `joint name` attribute is not found in the `unordered_set<std::string>`. Else we would need to rely on pointers for the return value.

luca-della-vedova:
Thanks for addressing all the feedback! Looks good to me

PR #118, opened 2020-04-17
Feature/teleport dispenser

aaronchongth:
* model names (instance names) now editable

* model names generated for SDF are now only appended with indices only if there are duplicates. Previously,
```
Table1
Table2
Chair3
Person4
Person5
```
now,
```
Table
Table2
Chair
Person
Person2
```
This allows uniquely named models with functions (eg. dispensers) to set their `guid` to their model names, so users can design model interactions from the traffic editor.

* added `TeleportDispenser` and `TeleportIngestor` thumbnails, they are the same as `RobotPlaceholder`

Let me know if you would prefer for this PR to wait for your merge over the next few days.


aaronchongth:
I was afraid you made more elaborate changes than that for other purposes, thanks for the quick response!

PR #119, opened 2020-04-28
Level name populated in Slotcar RobotState msg

Yadunund:
This PR addresses [issue 6](https://github.com/osrf/romi-dashboard/issues/4) for the romih dashboard.
To correctly initialize this `level_name` when the plugin is loaded, the plugin parses the `BuildingMap` msg as published by the `building_map_server` and identifies the level based on proximity of the spawned model to a waypoint on a level. Subsequently, the `level_name` is inferred from incoming `PathRequest` msgs which should take care of situations where the robot takes a lift to another level.


Yadunund:
Thanks! Setting the `level_name` in the world file would be ideal. Will work towards that goal!

PR #136, opened 2020-05-29
Fix model name parsing in world generation

methylDragon:
## Problem
When model paths are populated into the `.world` file specifications during `.world` generation, the model paths are populated by naively copying the thumbnail name that corresponds to the model.

Unfortunately, due to changes in how we are now specifying models (https://github.com/osrf/traffic_editor/pull/132, https://github.com/osrf/traffic_editor_assets/pull/2), this would result in an erroneous model path being specified (i.e., one that includes a namespace.)

## Solution
As such, the world generation pipeline will be edited to explicitly remove any namespaces if they exist. This should not change any user-facing workflows or interfaces, and is fully back-compatible with old-style thumbnail specifications.

Although due to https://github.com/osrf/traffic_editor_assets/pull/2 and https://github.com/osrf/traffic_editor_assets/pull/3 , all future thumbnails will be using the new-style thumbnail specifications with namespaces moving forward.

Additionally, the README has been updated to reflect the change in thumbnail default directory.

## Caveats
Note then that since we will be removing namespaces during world generation, this will prevent users from manually inserting namespace models into their `GAZEBO_MODEL_PATH`, but this is a **minor** issue as this is **not** the standard way to specify Gazebo models as it contradicts Gazebo's normal behaviour when downloading models (as well as `pit_crew`'s behaviour when downloading missing models during world generation.)

Additionally, note that the `model_name` attributes under each model in generated `building.yaml` files actually refer to the thumbnail name (and path), rather than the corresponding model.

E.g.
`building.yaml` : {model_name: **OpenRobotics/VendingMachine**, name: VendingMachine, static: true, x: 62.373, y: 73.121, yaw: -3.1416, z: 0}
Would imply that the corresponding:
- Thumbnail will be named **OpenRobotics/VendingMachine** and be stored in **OpenRobotics/VendingMachine**
- Model will be found in the **<model_dir>/VendingMachine/** path directory.

Notice that the model path does **not** contain the namespace of the author (in this case OpenRobotics.)

- CH3EERS!


methylDragon:
https://github.com/osrf/traffic_editor/pull/136/commits/9001f2dbd13345173980632e96cf179d88967c76 now substitutes missing non-namespaced thumbnails with namespaced ones, with a warning.

And https://github.com/osrf/traffic_editor/pull/136/commits/6b95ff500df57061bcfcc23886c7f80ca0e3343e reassigns the configured model_name so that it will save as that substituted namespace model_name moving forward.

Future work:
Implement an `author_name` attribute and change how the `model_list.yaml` is structured to reflect namespaces.

Will have to do it in as backcompatible a way as possible.

traffic_editor/gui/building_level.cpp
```diff
@@ -837,8 +837,38 @@ void BuildingLevel::draw(
...
+    if (pixmap.isNull()) {
+      // BACKWARDS COMPATIBILITY PATCH: Try again, but...
+      // Use the first namespaced thumbnail for a specified non-namespaced
+      // model with a warning instead. (Also modifies the model name inplace!)
+
+      for (auto &editor_model : editor_models)
+      {
+        // Check if non-namespaced name is a substring of namespaced name
+        // If yes, use it!
+        if (editor_model.name.find(model.model_name) != std::string::npos)
```
aaronchongth, line 849: We should compare the whole `editor_model.name`'s suffix with `model.model_name`, otherwise we might get cases of `table` getting paired with `OpenRobotics/cafe_table` instead of `OpenRobotics/table`

aaronchongth: for example,
```
[WARNING] Thumbnail table not found, substituting OpenRobotics/cafe_table instead!
(table will be saved as OpenRobotics/cafe_table!)

[WARNING] Thumbnail TrolleyBed not found, substituting OpenRobotics/SmartTrolleyBed instead!
(TrolleyBed will be saved as OpenRobotics/SmartTrolleyBed!)
```

methylDragon: Fixed in https://github.com/osrf/traffic_editor/pull/136/commits/1052bcb7e6d2bc057a0adcd13dbf4801f4b6e1a2 !

methylDragon:
Hopefully, I didn't miss out on any more edge cases, hmm..

methylDragon:
I think we should be ok now.

Any further work that will help get things stable will probably be involving thumbnail regeneration with the fixes this PR introduces.

I think we are good to merge.

methylDragon:
I don't have write permissions to merge this in, so I think you can go ahead if you can @aaronchongth

PR #137, opened 2020-05-29
Feature/add lift support

Kevinskwk:
Adding support for lifts. The lifts generated should be minimally functional but some manual works are still required as some features are not yet fully implemented.


building_map_tools/building_map/lift.py
```diff
@@ -10,33 +18,291 @@ def __init__(self, yaml_node, name):
...
+                self.door_names.append(door_names)
                 self.level_names.append(level_name)
-                self.door_names.append(door_name)
 
         self.doors = []
         if 'doors' in yaml_node:
             self.doors = self.parse_lift_doors(yaml_node['doors'])
 
+        # for wall generation
+        self.vertices = \
```
Yadunund, line 169: The name `vertices` is a bit misleading. For `front` and `back` the array represents the x-coordinate of the corner vertices while for `right` and `left`, it represents the y-coordinate of the vertices. (correct me if I'm wrong). The same conventions is used for `door.side`. Would switching to a list of tuples (x-coord, y-coord) for both places be more intuitive?

Kevinskwk: I intend to make each array represent the 1-dimensional positions of the endpoints on each side. It makes the process of sorting and picking endpoints of walls easier. I think switching to tuples of (x, y) is a bit redundant. Maybe changing the dictionary name to `end_points` will be better?

Kevinskwk:
Thanks for all your comments! I have finished cleaning up the lift sdf generation and refactoring the lift plugin. As I suffered a lot from using `unordered_map` with `unique_ptr`, I'm not 100% sure whether the lines of code I used to manipulate unique_ptr's are of good coding practice. Please kindly point them out if there're better approaches. Thank you!

Kevinskwk:
@Yadunund Thanks for reviewing for me and all your comments helping me improving the code. Glad that I can contribute to the main release!

PR #138, opened 2020-05-29
Ignition plugins and modularization of doors and slotcar

luca-della-vedova:
This PR adds ignition support for doors and robot vehicles and splits the building_gazebo_plugins package into three separate ones:

- building_sim_common: A library that contains all the functionality that is agnostic of the simulator being used.
- building_sim_gazebo: The gazebo plugins, depending on the common library.
- building_sim_ign: The ignition plugins, depending on the common library as well.

Since the plugins are moved from building_gazebo_plugins package to building_sim_gazebo downstream packages (and launch files) would have to change the package they are using for their plugins.


luca-della-vedova:
Yap! I tested as far as I could on the matching [rmf_demos branch](https://github.com/osrf/rmf_demos/tree/feature/add_ignition_support). Another pair of eyes is always appreciated

aaronchongth:
sure thing, I'll take a look at it. we can probably pull in `master` again and resolve any merge conflicts after #139 is merged

luca-della-vedova:
Feel free to merge #139 first and I'll port the changes

luca-della-vedova:
#139 changes ported

luca-della-vedova:
> it looks great! thanks for modularizing the plugins, one tiny recommendation would be to parameterize the topics instead of hardcoding, eg. `/door_states` and `/door_requests` in `door_common.cpp`, but that is not critical and should not hinder any merging. thanks again!

Do you mean changing to a class `static constexpr std::string` or make them actual variables? Who is in charge of defining their values then?

aaronchongth:
> Do you mean changing to a class `static constexpr std::string` or make them actual variables? Who is in charge of defining their values then?
Sorry I meant allowing them to be modified and defined in the the model SDF, for example.

luca-della-vedova:
> > Do you mean changing to a class `static constexpr std::string` or make them actual variables? Who is in charge of defining their values then?
> 
> Sorry I meant allowing them to be modified and defined in the the model SDF, for example.

Yea I guess it can be done in a second moment then (this PR doesn't really add any functionality). What would be the use case? Iirc rmf_core uses a global topic for avoid having too many topics and making discovery tricky.

luca-della-vedova:
Noticed (thanks @Yadunund !) that merging this PR introduces the necessity for every user to install the latest ignition from source which can be quite troublesome, especially if they only want to use gazebo classic.
Added a [change in CMakeLists](https://github.com/osrf/traffic_editor/pull/138/commits/aef49650d44199491feab6a7e2d8066d0c50084c) to avoid this and make the compilation quit early (without breaking the whole build) if the correct version of ignition is not detected.

PR #147, opened 2020-06-16
Fix not generating cropped image with namespace

MakinoharaShouko:
I opened [this issue about thumbnail generator](https://github.com/osrf/traffic_editor/issues/146) yesterday. I went through crop.py and found that when a namespace is used cv2.imwrite would not be able to generate the cropped image because a directory for the namespace needs to be created first, so I made some small changes to crop.py and added a mkdir for the author name

### DIFF
--- a/traffic_editor/thumbnail_generator/scripts/crop.py
+++ b/traffic_editor/thumbnail_generator/scripts/crop.py
@@ -135,7 +135,7 @@ if __name__ == '__main__':
         )
 
         output_filepath = os.path.join(
-            os.path.abspath(args.output_dir),
+            os.path.expanduser(args.output_dir),
             '{}.png'.format(model_name)
         )
 


methylDragon:
This looks good to me

Unrelated to this are a couple of other fixes pending merges to support namespaced models.

Do look out for them (or forward port the [standing PR](https://github.com/osrf/traffic_editor/pull/136))

CH3EERS!

methylDragon:
@codebot

The other PR is dependent on the models being available on Fuel and new thumbnails being generated for them.

That one might cause issues with  world generation when new namespaced models get used in traffic editor projects due to how the models get processed during world generation.

It's still pending testing after we have all our models on Fuel though. I'll hesitate merging that for now just in case something else breaks for any other current users of traffic_editor. (Lots of moving parts... :( )

methylDragon:
For now just the tedium of the process I think.

It seems to be quite the undertaking!

PR #149, opened 2020-06-17
process-flow sim plugin interface, and various other improvements

codebot:
This PR got a bit out of control. It's a few month's worth of various improvements:
 * various C++ style fixes throughout, to better follow `rmf_core` style
 * to prepare for release via ROS build farm, `traffic-editor` is a proper Colcon package now
 * restore previously-opened level on startup (for multi-level maps)
 * load background images in parallel (for multi-level maps)
 * plugin interface to allow creation of closed-source process-flow simulation plugins that run in a separate thread
 * video recording of process-flow simulations using OpenCV (optional dependency, builds fine without it)
 * only render arrows on lanes for unidirectional segments. The previous "double-direction arrows" were just too messy.
 * render lanes in a consistent Z stacking order, so that the composite colors of overlapping lanes are consistent
 * traffic graphs can now be individually toggled as shown/hidden in the `Traffic Maps` tab
 * change font and size of vertex names so they're easier to read
 * move most include files to a `include/traffic_editor` directory, so they can be installed for process-flow simulation plugins to find later
 * factor model drawing into Model class (general cleanup, a few other similar refactorings to try to reduce the size of the `editor.cpp` monolith)
 * factor model state (x, y, yaw, etc.) into a `ModelState` class for cleaner manipulation in process flow plugins


traffic_editor/package.xml
```diff
@@ -0,0 +1,15 @@
...
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>traffic_editor</name>
+  <version>0.0.0</version>
+  <description>traffic editor</description>
+  <maintainer email="morgan@openrobotics.org">Morgan Quigley</maintainer>
+  <license>Apache License 2.0</license>
+
+  <build_depend>ament_cmake</build_depend>
+
```
marcoag, line 11:
```suggestion
  <build_depend>libignition-plugin-dev</build_depend>
  <build_depend>libignition-common3</build_depend>
```

codebot: I agree we need some sort of mention here about `ignition-plugin` but I'm I'm not sure if it needs the `-dev` suffix?

PR #150, opened 2020-06-18
Update style.yaml

Yadunund:


Yadunund:
> Can `ament_uncrustify` output be printed into the GitHub action output? Maybe I'm looking in the wrong place, but it seems to just say this:

Thanks for catching that. It looks like `uncrustify` was having a hard time parsing a couple of blocks in the `project.cpp` file. Commit https://github.com/osrf/traffic_editor/pull/150/commits/6f7292a296506a06f0311b3007b901d1dea08e25 adds braces to a couple of `for` loops which seems to fix the issue. The output of `uncrustify` should now be printed.

PR #152, opened 2020-06-19
Implement using thumbnails from installed traffic_editor_assets ament package

methylDragon:
With the change in `traffic_editor_assets` in https://github.com/osrf/traffic_editor_assets/pull/5 , there is no longer any need to pull the repository on build! This means that you no longer need an internet connection to have a successful build.

Instead, simply build `traffic_editor_assets` and `traffic_editor` will take care of the rest, dynamically resolving the location of `traffic_editor_asset`'s thumbnail directory instead of using a hidden folder in your home directory!

## Testing
To test it, remember to remove the default preferences from `traffic_editor` in `~/.config/open-robotics` so that it will search for the default thumbnail path.

Then run `traffic-editor` and watch the console for its efforts in locating the `traffic_editor_assets` package (make sure the assets package is built and the workspace is sourced, of course!), and verify that the thumbnails could be found by trying to add a model.

### DIFF
--- a/traffic_editor/package.xml
+++ b/traffic_editor/package.xml
@@ -12,7 +12,9 @@
   <build_depend>traffic_editor_assets</build_depend>
   <build_depend>libignition-plugin-dev</build_depend>
   <build_depend>libignition-common3</build_depend>
-  
+
+    <exec_depend>traffic_editor_assets</exec_depend>
+    
   <export>
     <build_type>ament_cmake</build_type>
   </export>
--- a/traffic_editor/package.xml
+++ b/traffic_editor/package.xml
@@ -9,11 +9,8 @@
 
   <build_depend>ament_cmake</build_depend>
   <build_depend>ament_index_cpp</build_depend>
-  <build_depend>traffic_editor_assets</build_depend>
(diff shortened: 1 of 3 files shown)


traffic_editor/package.xml
```diff
@@ -8,6 +8,8 @@
   <license>Apache License 2.0</license>
 
   <build_depend>ament_cmake</build_depend>
+  <build_depend>ament_index_cpp</build_depend>
+  <build_depend>traffic_editor_assets</build_depend>
   <build_depend>libignition-plugin-dev</build_depend>
   <build_depend>libignition-common3</build_depend>
```
marcoag, line 15:
```suggestion

    <exec_depend>traffic_editor_assets</exec_depend>
    
```

`traffic_editor` will crash if the `traffic_editor_assets` package is not there at execution time...

methylDragon: Good catch!

methylDragon: *** 
https://github.com/osrf/traffic_editor/pull/152/commits/6eed9df290f445eb448ca881046d59816c5ab2de

Caught the exception instead and removed the dependency, printing with a warning

PR #159, opened 2020-06-26
consistent handling of model angles. Add GUI to transform all model yaw

### LINKED ISSUE #156: Rotating Thumbnails
* Attempting to rotate model thumbnails always places them -90 degrees offset from what was desired

rotation_offset

codebot:
Fix #156  by migrating towards the "typical" Cartesian angle convention of 0 radians = east, pi/2 radians = north. Because Qt uses the "geographic" convention of zero degrees = north, 90 degrees = east, we've flip-flopped on this a bit.

To deal with the past, there is now a GUI dialog in `Edit->Transform` that lets you enter an angle offset to "fix" maps that have the opposite convention. `Edit->Transform...` will instantly "fix" the `rmf_demos` office and airport maps if you enter `1.5708` :sparkles:

### DIFF
--- a/building_map_tools/building_map/model.py
+++ b/building_map_tools/building_map/model.py
@@ -39,7 +39,7 @@ class Model:
         pose_ele = SubElement(include_ele, 'pose')
         x, y = transform.transform_point((self.x, self.y))
         z = self.z
-        yaw = self.yaw + 1.5707 + transform.rotation
+        yaw = self.yaw + transform.rotation
         pose_ele.text = f'{x} {y} {z} 0 0 {yaw}'
 
         static_ele = SubElement(include_ele, 'static')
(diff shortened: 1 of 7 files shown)


Yadunund:
Hi @codebot ,

I did a clean build of this branch and loaded the `office` map from `rmf_demos` in traffice_editor. The thumbnails were rotated by 90deg as expected so I applied the `Edit->Transform` option as suggested. All the thumbnails rotated to desired orientations as seen in the image below. However, building the map leads to the objects having 90deg offsets in `Gazebo`.

image

Yadunund:
> However, building the map leads to the objects having 90deg offsets in `Gazebo`.

I think the culprit is this hardcoded `1.5707` in this line
https://github.com/osrf/traffic_editor/blob/a76ba8053689949aa99b7cc2523b1b8ae15234b4/building_map_tools/building_map/model.py#L42

Removing that constant and rebuilding has produced desirable results 
image

PR #161, opened 2020-06-29
Add missing includes and namespace usages for math functions

gbiggs:
g++-9 and/or C++11/14 require the `cmath` header and `std::` namespace on various math functions used in the traffic editor GUI. g++8 probably wasn't being strict about these. Focal uses g++9 by default so they need to be fixed.


gbiggs:
Actually I haven't checked. I assumed CI would check that for me. I'll try it out.

gbiggs:
The changes work fine on Bionic.

PR #164, opened 2020-06-29
Handle for missing directories in pit_crew

methylDragon:
https://github.com/osrf/traffic_editor/issues/163 highlighted an issue when a user's Gazebo model directory does not exist.

This PR fixes that issue.

For output directories, it creates them.
For parsed directories, it logs a warning instead of throwing an assertion, and follows behaviour as if the parsed directory were empty instead of non-existent.


methylDragon:
Merging to `master` instead

PR #166, opened 2020-06-30
Use C++17 standard to match SDFormat9 requirement

gbiggs:
SDFormat version 9 [requires C++17](http://sdformat.org/tutorials?tut=roadmap). It uses `std::variant` which is only available from C++17 onwards. This PR changes the C++ version for the `building_gazebo_plugins` package.

This should be held until after the release.

### DIFF
--- a/building_gazebo_plugins/CMakeLists.txt
+++ b/building_gazebo_plugins/CMakeLists.txt
@@ -9,7 +9,7 @@ if(NOT CMAKE_C_STANDARD)
   set(CMAKE_C_STANDARD 99)
 endif()
 
-# Default to C++14
+# Default to C++17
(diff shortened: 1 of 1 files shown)


Briancbn:
Hi,
Although the build passed in eloquent, the **rmf_demos** run will fail if you simply change to this.

I would suggest the following after `find_package(gazebo)`
```cmake
if (NOT ${GAZEBO_VERSION} VERSION_LESS "11.0.0")
  set(CMAKE_CXX_STANDARD 17)
endif()
```

gbiggs:
Which package is that in?

Yadunund:
Closing as the branch has been renamed to `foxy` as per #194

PR #168, opened 2020-07-01
so embarrassing. pretend this never happened.

codebot:
Uhhh so... yeah. The double-ampersand doesn't mean what I thought it did :man_facepalming:

Nobody was using this function anyway; I was attempting to use it in a plugin and then realized that I don't know C++.


gbiggs:
rvalue references are great for moving things, it's so efficient! Until you discover you still want your copy of the object...

PR #179, opened 2020-07-06
Formatting Fix

Yadunund:
Modified formatting of C++ files to align with our RMF codestyle. Given the large number of lines that have changes, it would be great to have a more people test the build and functioning of everything :bow:

### DIFF
--- a/building_gazebo_plugins/src/door.cpp
+++ b/building_gazebo_plugins/src/door.cpp
@@ -21,9 +21,9 @@ public:
   bool _debuggable;
 
   Door(const bool debuggable,
-      const gazebo::physics::JointPtr& joint,
-       const MotionParams& params,
-       const bool flip_direction = false)
+    const gazebo::physics::JointPtr& joint,
+    const MotionParams& params,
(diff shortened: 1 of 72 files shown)


gbiggs:
Does RMF have a declared coding style? If not, is it better to do this massive changeset to align with RMF, or to do a massive changeset to align both with a declared coding style such as the one used by ROS 2?

Yadunund:
> Does RMF have a declared coding style? If not, is it better to do this massive changeset to align with RMF, or to do a massive changeset to align both with a declared coding style such as the one used by ROS 2?

@gbiggs, for RMF related projects, the team has adopted a set of style-related rules for C++ files which slightly differ from the default ROS 2 style guide. We're using `ament_uncrustify` as our C++ linter with rules configured in this [file](https://github.com/osrf/rmf_core/blob/master/rmf_utils/test/format/rmf_code_style.cfg). So far `rmf_core`, `rmf_demos` and `rmf_schedule_visualizer` repositories have been linted with this file. I'm currently in the process of formalizing these rules into a document for contributors.

PR #188, opened 2020-07-13
handle situation when lifts key is not present

### LINKED ISSUE #187: building map generator: don't crash if building doesn't define lifts key
Should be a quick fix, just need to wrap this loop with a conditional on `yaml_node[lifts]`:

https://github.com/osrf/traffic_editor/blob/master/building_map_tools/building_map/building.py#L34

Kevinskwk:
Simple fix to handle buildings without `lifts` key defined.

Fixes #187

### DIFF
--- a/building_map_tools/building_map/building.py
+++ b/building_map_tools/building_map/building.py
@@ -31,14 +31,15 @@ class Building:
         self.transform_all_vertices()
 
         self.lifts = {}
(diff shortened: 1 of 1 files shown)


luca-della-vedova:
Can you do a quick rebase on the latest master to see if it fixes style? Just to make sure we are not introducing any issue with the increased line length.

PR #191, opened 2020-07-16
New traffic editor thumbnail generator

tanyouliang95:
A new thumbnail generator for traffic_editor!! :tada: 

Development is referred to [ModelPropShop](https://github.com/osrf/gazebo/blob/gazebo11/plugins/ModelPropShop.cc) and the current [thumbnail_generator](https://github.com/osrf/traffic_editor/tree/master/traffic_editor/thumbnail_generator) implementation. 

Compared to the previous implementation, this won't require ros1 as a dependency and more simplified. User will just need to run a single line of command to generate the thumbnail:  `gzserver -s libthumbnail_gen.so empty.world ${INPUT_SDF_FILE} ${OUTPUT_FOLDER}`. To run multiple models, run the provided py script. 

This seeks to replace the current `thumbnail_generator`. Let me know what needs to be added. If goes well I will remove the current `thumbnail_generator` and move those utils script to `traffic_editor/script`


building_gazebo_plugins/src/thumbnail_generator.cpp
```diff
@@ -0,0 +1,277 @@
...
+#include "gazebo/msgs/msgs.hh"
+#include "gazebo/transport/transport.hh"
+#include "gazebo/sensors/SensorsIface.hh"
+#include "gazebo/common/Plugin.hh"
+#include "gazebo/rendering/rendering.hh"
+#include "gazebo/util/system.hh"
+
+#define IMG_SIZE 4000
+#define CAM_HFOV 0.08
+#define CAM_HEIGHT 200
```
aaronchongth, line 32: It would be great if we could parameterize these macros as well, and have them as default values, unless specified in command line. I would suggest writing an arg parser for all the arguments necessary, including the model paths and output directories

building_gazebo_plugins/src/thumbnail_generator.cpp
```diff
@@ -0,0 +1,277 @@
...
+  /////////////////////////////////////////////
+  virtual ~ThumbnailGenerator()
+  {
+    rendering::fini();
+  }
+
+  /////////////////////////////////////////////
+  void Load(int _argc, char** _argv)
+  {
+    if (_argc != 6)
```
aaronchongth, line 49: writing an arg parser with proper flagging like `--model` and `--output` might be a cleaner way to handle more arguments in the future.

traffic_editor/scripts/generate_thumbnails.py
```diff
@@ -0,0 +1,63 @@
...
+        parser.add_argument(
+                'models_dir', default='~/.gazebo/models',
+                help='Directory of models to be added')
+        parser.add_argument(
+                'model_list', default='../test/model_list.yaml',
+                help='Path of model_list.yaml')
+        parser.add_argument(
+                'output_dir',
+                help='Directory where the output images will be saved')
+        args = parser.parse_args(argv[1:])
```
aaronchongth, line 35: just a note, that if you manage to add the arg parser in the plugin, you will also be able to propagate the arguments to this script as well

tanyouliang95:
Thanks for the review! Cleaned the subdir and added argparsing for both scripts.  :+1:

PR #194, opened 2020-07-22
Foxy

Yadunund:


gbiggs:
That API does not exist in Eloquent.

gbiggs:
Yes, I think that's probably doable and manageable. Is there a globally-available define available for the in-use ROS distro?

gbiggs:
It compiles for me on Ubuntu 20.04/Foxy

gbiggs:
It turns out that Ignition is supposed to provide the version defines, but it appears to be broken.

```
#define IGNITION_MATH_MAJOR_VERSION 
#define IGNITION_MATH_MINOR_VERSION 
#define IGNITION_MATH_PATCH_VERSION 

#define IGNITION_MATH_VERSION "6.4"
#define IGNITION_MATH_VERSION_FULL "6.4.0"
```

(From `/usr/include/ignition/math6/ignition/math/config.hh`)

gbiggs:
It appears to be a bug in Ignition Math (other Ignition libraries have correct defines). I think we should merge with the Gazebo check for now and change it when the Ignition Math bug is fixed.

gbiggs:
When this goes in and a new release appears, we can fix it.
https://github.com/ignitionrobotics/ign-math/pull/151

Yadunund:
We should update the Github actions to use the Ubuntu 20.04/foxy containers in a separate PR :+1:

PR #195, opened 2020-07-23
cleanups for nested model in ignition

Kevinskwk:
Cleanups for unused code in prepare of nested model support for ignition.
Should be ready to get merged when https://github.com/ignitionrobotics/ign-gazebo/pull/258 is merged for `ign-gazebo4`.


luca-della-vedova:
A similar but not quite the same fix has been bundled in #483.
Specifically, we can't naively remove the nesting as we planned to because Gazebo does not allow nested model joints to refer to their parents. What we have to do instead, is move the joints up in the hierarchy to be in the parent model and have them refer to the child model instead. Closing this

PR #200, opened 2020-07-28
Configurable texture and transparency for walls

tanyouliang95:
User will now be able to specify texture and alpha (assuming a glass wall!) for each wall in `traffic_editor`. A simple screenshot to illustrate: 

gz_screeenshot

During generation, `building_map_generator` will generate separate meshes and `obj` files, corresponding to the params (e.g. texture_name and alpha). Note that the wall texture is not in a "tile" format yet.


tanyouliang95:
Thanks for the review! Yes, we will upgrade from 1D wall texture to 2D texture when there's a request. Tested `rmf_demo_maps` with this building map generation. Good to go. :+1:

PR #201, opened 2020-07-29
Support for adding and recognizing lift waypoints for multi-level navigation

Kevinskwk:
This PR involves two features:

- Checking the geometry of vertices and lifts. Automatically setting "lift: [lift name]" properties for waypoints inside lifts. This supports the multi-level nav-graph connection in https://github.com/osrf/rmf_core/pull/139

- Adding a button in traffic_editor GUI for users to generate vertices aligned at the center of the lift on each level.

Thanks for reviewing!


Kevinskwk:
Thanks for the feedback and suggestion! I've chosen a middle ground to keep the button, but it now won't create duplicated vertices. Vertices generated by lifts now have the `"belongs_to"` parameter indicating which lift is it generated by. The lift will then track these vertices and update their coordinates (or re-generate them if missing) whenever the button is clicked or the lift is moved.

Please let me know if any further changes are required.

Kevinskwk:
Sorry for those small errors. Fixed!

Kevinskwk:
Sure thing, I'm changing the parameter name.

Kevinskwk:
Before I aligned the waypoints based on the coordinates in meters displayed in the bottom-right properties table. Now they should align at the lift center on all levels

PR #203, opened 2020-08-01
Fixing lift doors not updated after renaming the levels

MakinoharaShouko:
In the current traffic editor, when the user renames a building level, the lift doors do not update and the lift doors on the old levels still remain in the .building.yaml file. This update should fix this problem. I have tested on my own computer and everything looks fine.

### DIFF
--- a/traffic_editor/gui/building_level_dialog.cpp
+++ b/traffic_editor/gui/building_level_dialog.cpp
@@ -19,7 +19,7 @@
 #include <QtWidgets>
 
 
-BuildingLevelDialog::BuildingLevelDialog(BuildingLevel& _level, Building* _building)
+BuildingLevelDialog::BuildingLevelDialog(BuildingLevel& _level, Building& _building)
 : building_level(_level), building(_building)
 {
   ok_button = new QPushButton("OK", this);  // first button = [enter] button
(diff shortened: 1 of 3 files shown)


MakinoharaShouko:
Just changed the pointers to references.

traffic_editor/gui/building_level_dialog.cpp
```diff
@@ -180,7 +180,14 @@ void BuildingLevelDialog::ok_button_clicked()
       "Name must not be empty");
     return;
   }
+  auto original_name = building_level.name;
   building_level.name = name_line_edit->text().toStdString();
+  for (size_t i = 0; i < building.lifts.size(); i ++) {
```
codebot, line 185: I believe the C++ style checker on this repo will flag this (it's set up for block-braces, not cuddle-braces). Do you have GitHub Actions enabled for your fork of this repo? If so, then it should start running the build-test and style-test checks for you.

codebot: Because the PR is coming from a fork, GitHub isn't running the checks against this PR

MakinoharaShouko:
Now my checkers are both running and pass!

MakinoharaShouko:
> Nice! How did you get them to run on your fork?

To be honest I am not sure. I ran the checker locally once and then it started working when I pushed

PR #205, opened 2020-08-04
Use door name in plugin name to prevent node name clashes

gbiggs:
`gazebo_ros` creates nodes using the `name` value of the `plugin` element. This PR makes the SDF generator use the name of a door in its plugin name to prevent node name clashes when more than one door is present in the world.

Signed-off-by: Geoffrey Biggs <gbiggs@killbots.net>

### DIFF
--- a/building_map_tools/building_map/doors/double_sliding_door.py
+++ b/building_map_tools/building_map/doors/double_sliding_door.py
@@ -22,7 +22,7 @@ class DoubleSlidingDoor(Door):
             options)
 
         plugin_ele = SubElement(self.model_ele, 'plugin')
-        plugin_ele.set('name', 'door')
+        plugin_ele.set('name', f'door_{self.name}')
         plugin_ele.set('filename', 'libdoor.so')
         plugin_params = {
           'v_max_door': '0.2',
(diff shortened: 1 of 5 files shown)


luca-della-vedova:
I'll check on the lift / ignition side of things in the coming days, I wonder if by doing this we could remove the [hack](https://github.com/osrf/traffic_editor/blob/master/building_sim_plugins/building_ignition_plugins/src/door.cpp#L73-L75) currently used in the ignition door plugin to manually assign the plugin name, which should also get us more consistency in naming between the two simulators

luca-della-vedova:
OK this broke door plugins in ignition, investigating (will push directly to this branch when I find the fix).

luca-della-vedova:
The errors printed look like the following (error extracted from the office world)

```
[ign-8] [Err] [SystemLoader.cc:93] Failed to load system plugin [door_main_door] : could not instantiate from library [libdoor.so] from path [/home/luca/demos_ws/install/building_ignition_plugins/lib/libdoor.so].
```

The issue is that the [plugin loader in ignition gazebo](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/SystemLoader.cc#L90) uses the `name` property to look for the plugin inside the library, the plugin name (or alias) is defined in the [plugin's source file](https://github.com/osrf/traffic_editor/blob/master/building_sim_plugins/building_ignition_plugins/src/door.cpp#L156) hence I don't think there is an easy fix to make this work with ignition.

I wonder if it is possible to pass a custom node name to gazebo_ros instead?

gbiggs:
It didn't appear to be possible, but I'll take a deeper look.

luca-della-vedova:
Closing for inactivity and to prepare the repo migration

PR #216, opened 2020-08-11
Update lift display

MakinoharaShouko:
In the current version of traffic editor, lifts are displayed on all the floors. However, sometimes there are lifts that only visit certain levels. Displaying lifts on levels they cannot reach does not make sense and overlaps with models as well. To make things more clear, I added the feature that lifts are only displayed on the levels they stop.

Additionally, I also added the feature to only display the doors open on the current level mentioned in the comment of lift.cpp


MakinoharaShouko:
That's true. In fact I am now dealing with a building that some lifts only go through part of all the levels. I think then there should be some other way to distinguish between not stopping and not able to reach

MakinoharaShouko:
Just changed the rendering of non-stopping lifts to red (maybe pink would be more appropriate).

I used gray at first but thought that may not be obvious enough, especially when there is a hole polygon.

The implementation of selecting reachable floors would be a lot of work and I think it should go to another PR

PR #218, opened 2020-08-19
Fred/crowd sim plugin

FloodShao:
Migrating the crowd simulation plugin into building_sim_plugins package.
Please include the import the menge lib to src directory when compiled the plugins, otherwise, the crowd simulation plugins will not be compiled.
The current using menge lib is as following.
```
repositories:
    meng_core:
        type: git
        url: https://github.com/FloodShao/menge_core.git
        version: master
```
To run the crowd simulation example. 
For gazebo:
1. go to .../building_gazebo_plugins/test
2. `ros2 launch office.crowd_simulation.launch.xml`

For Ign:
1. go to ../building_ignition_plugins/test
2. in the launch file `office.launch.xml`, you might need to add the model path for common models in <https://github.com/FloodShao/traffic_editor/blob/fred/crowd_sim_plugin/building_sim_plugins/building_ignition_plugins/test/office.launch.xml#L7>. 
3. `ros2 launch office.launch.xml`

There is an additional issue for ignition plugin:
Current test example for ignition plugin does not assign magni1 and magni2 as external agents, though they are imported in the world. The reason is when processing the crowd simulation initialization, the magni model (plugin) might not be imported in the world yet, which makes the crowd simulation plugin failed to be initialized. This situation isn't considered before because previous demo didn't import other plugin.


luca-della-vedova:
As a high level comment I would remove all the testing files if we want to merge this. A strategy you could try is move all the testing files to a separate branch (i.e. called `crowd_sim_plugin_test).
We can do high level tests using that branch and merge this one when it looks good. This should make the diff much smaller and easier to manage

luca-della-vedova:
OK got it up and running! I was pondering if we can do something to make the actors look more natural when they reach a waypoint, basically what happens now is that they stop playing an animation at whatever point in time when they reach their destination (example below):

image

I can see at least two ways to make it look better:
1. Probably easier, reset the animation time to a constant (i.e. 0?) where they look standing
2. Better but maybe harder, we can use a different animation for walking compared to "being at waypoint". For example I think the actor you are using has a stand.dae file with an animation for standing in place. The nice thing about this would be that in the future we can change it to arbitrary animations, for example we can have actors perform actions at their destinations.

Do you think this would be possible?

FloodShao:
I think I have achieved using the first method, but not sure it's satisfied. as follows:
image
The stand.dae actually does not have a stand position keyframe. The whole loop walks 1m so I tried to get the key frame when the two legs are together at 0.25 m. 

This code is pushed to fred/crowd_sim_plugin_test branch

luca-della-vedova:
I think it's OK, it seems it's not so immediate to switch among different animations at runtime, we can leave this as a future feature.

building_sim_plugins/building_gazebo_plugins/include/crowd_simulator.hpp
```diff
@@ -0,0 +1,142 @@
...
+  {
+    this->_tasks.remove(task);
+  }
+}
+
+template<typename TaskType>
+size_t TaskManager<TaskType>::GetTasksCount()
+{
+  return this->_tasks.size();
+}
```
luca-della-vedova, line 76: Is this TaskManager logic necessary? The ignition plugin seems to do fine without it?

building_sim_plugins/building_gazebo_plugins/include/crowd_simulator.hpp
```diff
@@ -0,0 +1,142 @@
...
+  void Load(gazebo::physics::WorldPtr world, sdf::ElementPtr sdf) override;
+
+private:
+  gazebo::physics::WorldPtr _world;
+  gazebo::event::ConnectionPtr _updateConnectionPtr;
+  gazebo::common::Time _lastAnimTime;
+  gazebo::common::Time _lastSimTime;
+  size_t _objectsCount;
+
+  mutable bool _initialized = false; //avoid optimization from compiler
```
luca-della-vedova, line 102: I very rarely see mutable being used, why is it necessary?

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+ignition::math::Pose3d Convert(const crowd_simulator::AgentPose3d& agent_pose){
+  return ignition::math::Pose3d(
+    agent_pose.X(), 
+    agent_pose.Y(), 
+    agent_pose.Z(), 
+    agent_pose.Pitch(), 
+    agent_pose.Roll(), 
+    agent_pose.Yaw());
+}
```
luca-della-vedova, line 494: Could these functions be moved to the common library?

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
```diff
@@ -0,0 +1,499 @@
...
+  auto scaleTrans = ignition::math::Matrix4d::Identity;
+  auto actorScale = actorPtr->Scale();
+  scaleTrans.Scale(actorScale.X(), actorScale.Y(), actorScale.Z());
+  rootAnimTrans = scaleTrans * rootAnimTrans;
+  return rootAnimTrans.Pose();
+}
+
+
+//============================================
+bool CrowdSimulatorPlugin::_LoadParams(const sdf::ElementPtr& sdf)
```
luca-della-vedova, line 225: I believe all the sdf parsing could also be moved to the common library, since both ignition and gazebo parse the same `<plugin>` tag

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
```diff
@@ -0,0 +1,499 @@
...
+      }
+    }
+  }
+
+  return true;
+}
+
+
+//============================================
+bool CrowdSimulatorPlugin::_LoadModelInitPose(
```
luca-della-vedova, line 398: Same as the above, can be moved to the common library?

luca-della-vedova (changes requested):
Just a quick review, I believe we can further reduce the size of this PR by moving as much logic as possible to the common library and reducing code duplication between the ignition and the gazebo plugins.
There might be some additional functions I missed but I pointed out the ones I could find from a quick glance.
Ideally we would reach something similar to the [door plugin](https://github.com/osrf/traffic_editor/blob/master/building_sim_plugins/building_gazebo_plugins/src/door.cpp), which is super simple.

building_sim_plugins/building_gazebo_plugins/include/crowd_simulator.hpp
FloodShao: Not quite necessary, I can change it using traverse all the actors

building_sim_plugins/building_gazebo_plugins/include/crowd_simulator.hpp
FloodShao: Oh, this is not necessary. this is written in the early stage when I was reading the "clean code for c++". Now I'm sure this is not necessary.

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
FloodShao: Yes, this can be done.

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
FloodShao: Yes, this can be done

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
FloodShao: The purpose of putting the conversion in here is I want the common library independent with ignition lib. Because the gazebo plugin and the ignition plugin are using different version of ign lib. Like other plugins are using eigen as the connection. But other plugins are one-way connection. crowd_sim plugin has two-way for external_agent and other agent. So I think if the gazebo plugin and ign plugin are continuing use different ign lib versions, this function might need to keep, separately.... though I don't know for sure.

luca-della-vedova: I think you can do a templated function, similar to what is done in [slotcar_common](https://github.com/FloodShao/traffic_editor/blob/master/building_sim_plugins/building_plugins_common/include/building_sim_common/slotcar_common.hpp#L19-L26), then the header wouldn't have an ignition dependency

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+  double Pitch() const;
+  double Roll() const;
+  double Yaw() const;
+
+  double& X();
+  double& Y();
+  double& Z();
+  double& Pitch();
+  double& Roll();
+  double& Yaw();
```
aaronchongth, line 42: I would recommend using the Setters that you have defined below for any assignment, instead of giving mutable references.

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+  void X(const double& x);
+  void Y(const double& y);
+  void Z(const double& z);
+  void Pitch(const double& pitch);
+  void Roll(const double& roll);
+  void Yaw(const double& yaw);
```
aaronchongth, line 49: Normally we only use `const &` for complex types, and since we know we are only handling primitive types here, it is fine to pass by value, `void Yaw(double yaw);`

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+//===============================================================
+/*
+* class AgentPose3d
+* replace the common interface of ignition::math::Pose3d, which might apply with different version in different plugins
+*/
+class  AgentPose3d{
+
+public:
+  AgentPose3d();
+  AgentPose3d(double& x, double& y, double& z, double& pitch, double& roll, double& yaw);
```
aaronchongth, line 26: Since the constructor's implementation only copies the values of the primitive types, you should just pass by value

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
FloodShao: thanks for the notice, will modify it.

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
FloodShao: thanks for the comments on `AgentPose3d`. will modify it.

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+  {
+
+    _behaviorFile = this->_ResourceFilePath(_behaviorFile);
+    _sceneFile = this->_ResourceFilePath(_sceneFile);
+
+    if (this->_LoadSimulation())
+    {
+      this->initialized = true;
+    }
+    assert(this->initialized);
```
aaronchongth, line 88: I might be missing something, but it looks like the `initialized` boolean doesn't serve any specific purpose here. 

For ensuring everything is loaded or constructed correctly before proceeding with the simulation plugin, you can consider using factory methods, https://refactoring.guru/design-patterns/factory-method/cpp/example. It is currently used extensively in `rmf_core` as static `make` functions, for example,
https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/Adapter.hpp#L36-L80
https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Adapter.cpp#L215-L242

This will allow you to propagate a failure all the way to the plugin and fail the plugin gracefully, instead of needing asserts or boolean `initialized` checks.

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+    std::string animation;
+    double animationSpeed;
+    
+    //for ignition
+    std::string modelFilePath;
+  };
+
+  //Create a new record and returns a reference to the record
+  template<typename... Args>
+  Record* Emplace(Args&& ... args){
```
aaronchongth, line 139: I would recommend not using variadic templates for this function, as it is currently only called to create a pair. The calls to this function, seem to be passing in new raw pointers,
https://github.com/FloodShao/traffic_editor/blob/fred/crowd_sim_plugin/building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp#L282
https://github.com/FloodShao/traffic_editor/blob/fred/crowd_sim_plugin/building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp#L154

It would be more ideal to let this method handle the creation of the object `Record`, so that you can delete it in the destructor of `ModelTypeDatabase`, currently this object is leaking memory as the allocated memory is not deleted. Do consider using smart pointers to prevent that.

Another way around this is to change the `unordered_map` to hold `<std::string, Record>`, instead of the pointer as value, and `Emplace` returns a mutable reference to the `Record`

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+    auto pair = this->_records.emplace(std::forward<Args>(args)...); //return pair<iterator, bool>
+    assert(pair.second);
+    return pair.first->second;
+  }
+
+  //Get the total number of actors
+  size_t Size();
+
+  Record* Get(const std::string& typeName);
```
aaronchongth, line 148: As mentioned above, this method could also return a smart pointer to the record, or a `const &`

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+private:
+  std::unordered_map<std::string, Record*> _records;
```
aaronchongth, line 151: As mentioned above, this object is currently leaking memory as it is not deleting all the allocated memory.

FloodShao: Yep, realized that just now. changing to smart pointer

building_sim_plugins/building_plugins_common/src/crowd_simulator_common.cpp
```diff
@@ -0,0 +1,310 @@
...
+  std::cout << "Error in provided navmesh. Menge simulator initialized false." << std::endl;
+
+  return false;
+}
+
+//============================================
+
+ModelTypeDatabase::Record* ModelTypeDatabase::Get(const std::string& typeName)
+{
+  if (this->_records.find(typeName) == this->_records.end())
```
aaronchongth, line 90: A minor optimization that can be done here,
```
auto it = this->_records.find(typeName);
if (it == _records.end())
{
  return nullptr;
}
return it->second;
```

luca-della-vedova: Now that I look at it it should also be possible to make the function const by adding @aaronchongth optimization and removing the square bracket operator?

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
FloodShao: emm, `MengeHandle` class is a wrap up for the actual menge simulator on the purpose of only exposing the agent position related function. the _sim variable pointing to the actual menge simulator. After that, the CrowdSimInterface will use MengeHandle to access the agent position. `initialized` in Menge will be checked in CrowdSimInterface.

building_sim_plugins/building_plugins_common/src/crowd_simulator_common.cpp
FloodShao: Good idea. So the [] operator is usually not preferred? I should use iterator as much as possible right?

building_sim_plugins/building_ignition_plugins/include/crowd_simulator.hpp
```diff
@@ -0,0 +1,83 @@
...
+    // map for <model_name, entity_id> contains external and internal agents
+    std::unordered_map<std::string, ignition::gazebo::Entity> _entityDic;
+
+    std::shared_ptr<crowd_simulator::ModelTypeDatabase> _modelTypeDBPtr;
+    std::shared_ptr<crowd_simulator::CrowdSimInterface> _crowdSimInterface;
+    std::shared_ptr<rclcpp::Node> _nodePtr;
+    std::shared_ptr<ignition::transport::Node> _transportNodePtr;
+
+    volatile bool _pluginInitialized = false; // disable compiler optimization
+    volatile bool _agentInitialized = false;
```
aaronchongth, line 57: It looks like the whole plugin is going to run on a single thread, so there should be no need for the `volatile` keyword.

building_sim_plugins/building_plugins_common/src/crowd_simulator_common.cpp
aaronchongth: I will say it is still highly dependent on the use-case, @luca-della-vedova yes the method can be a `const` qualified, good catch

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
aaronchongth: Could you point me to the code that checks for it in `CrowdSimInterface`? I can't seem to find it.

aaronchongth (changes requested):
The simulation is working like a charm! Thanks for all the hardwork! This is going to be awesome once we iron out all the changes.

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
FloodShao: sorry, this `initialized` variable is not checked in `CrowdSimInterface`. Yeah, I should use the factory pattern. will modify that.

building_sim_plugins/building_ignition_plugins/include/crowd_simulator.hpp
FloodShao: Thanks, noted!

FloodShao:
Refactor the original crowd simulation plugin structure. Changes are follow:
1. Move the sdf loading part into 'common' interface
2. Move modeltype initialize part into 'common' interface
3. Use template function for passing actor pose between menge and plugin, to address the different ignition::math::Pose3d version problem.
4. Fixing original ignition plugin waiting agents loading problem.
5. Other code structure problems.

FloodShao:
To test the refactored crowd_sim_plugin, a new test branch is created under `fred/crowd_sim_plugin_refact_test`
https://github.com/FloodShao/traffic_editor/tree/fred/crowd_sim_plugin_refact_test/building_sim_plugins/crowd_sim_plugin_test

The test files are tested under the latest rmf_core structure, with tinyRobot (under both gazebo and ignition). 
Please use ignition version as https://github.com/osrf/rmf_demos/blob/feature/add_ignition_support/docs/ign-dome.yaml 
Take note the previous bug in Actor.cc is locally fixed in for sdf10.

luca-della-vedova:
Thanks for the fixes! The main issue I find now is in the code style, generally we have the [rmf guidelines](https://github.com/osrf/rmf_core/blob/master/docs/contributing.md) for contributing code that we try to stick to (for example function names should be `snake_case`). I think for consistency with the rest of the code base it would be good to follow the same guidelines, as well as applying the rmf linter.

building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+  std::shared_ptr<ModelTypeDatabase> _modelTypeDBPtr;
+  rclcpp::Logger logger() const;
+  void init_ros_node(const rclcpp::Node::SharedPtr node);
+
+  template<typename SdfPtrT>
+  bool readSDF(SdfPtrT& sdf);
+
+  bool initCrowdSim();
+  bool isInitialized() const;
```
aaronchongth, line 160: it looks like this function is not used anywhere, and since `initCrowdSim` returns a `bool` to determine the initialization state, there is no real use for the member `_initialized` as well

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
```diff
@@ -0,0 +1,273 @@
+#include <memory>
+#include <unordered_set>
+#include <regex>
+#include <cstdlib>
+
+#include <ignition/math/Pose3.hh>
+
+#include "crowd_simulator.hpp"
```
aaronchongth, line 8: We normally use double quotations to include header files using relative paths, so this means "In the same folder", so you can move the header files into the `src` instead of `include`. 

Header files that hold public API, to be used by downstream packages, are normally the ones put into the `include` folder, for plugin libraries we are sure that no downstream packages will be using any of the declared public API, hence all the source files for plugins are normally found in `src` and none in `include`, for example `door.cpp`, `lift.cpp`.

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
```diff
@@ -0,0 +1,273 @@
...
+void CrowdSimulatorPlugin::_UpdateInternalObject(double deltaTime, double deltaSimTime,
+  const crowd_simulator::AgentPtr agentPtr,
+  const gazebo::physics::ModelPtr modelPtr,
+  const crowd_simulator::ModelTypeDatabase::RecordPtr typePtr)
+{
+
+  if (!agentPtr)
+  {
+    RCLCPP_ERROR(_crowdSimInterface->logger(), "Null agentPtr when update Object!");
+    assert(agentPtr);
```
aaronchongth, line 109: Is there a reason the `assert` is in this check? Also do you intend to have the code continue executing even if the pointer is invalid? My suggestion would be to check for `nullptr` and `return`,
```
if (!agentPtr)
{
  RCLCPP_ERROR(_crowdSimInterface->logger(), "Null agentPtr when update Object!");
  return;
}
```

Or if this check is really for debugging purposes, maybe you can get away with only using a single `assert` call instead of an `if` case,
```
assert(agentPtr);
assert(modelPtr);
```

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+#include <ignition/math/Pose3.hh>
+
+#include <ignition/gazebo/components/Name.hh>
+#include <ignition/gazebo/components/Model.hh>
+#include <ignition/gazebo/components/Actor.hh>
+#include <ignition/gazebo/components/Pose.hh>
+#include <ignition/gazebo/components/Static.hh>
+
+#include "crowd_simulator.hpp"
```
aaronchongth, line 14: Similar to above, we can consider shifting the header into `src` since we are not exposing any public API

aaronchongth (changes requested):
Just a few more minor changes and we will be ready to merge, thanks for incorporating most of the comments in the previous review! 

2 more things you can do, is to add proper licensing information into all the source files, we use Apache 2.0, you can find them all throughout `rmf_core`, and might just need to change the year to `2020`.

Linting will also be needed. You can check out how it is done through this github action, https://github.com/osrf/rmf_core/blob/feature/style_action/.github/workflows/style.yaml

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
FloodShao: Thanks for pointing out. Yep I remember I wrote this assert in the early stage to abort the plugin once an invalid pointer is found, which I think `assert()` is doing the check, and if found false, assert will call `abort()`. Just return from this function might just get a log and the rest code is continuing. 
Is there any way to do it elegantly? maybe I should use `exit()` ?

aaronchongth (approved):
Thanks for making the changes for all the new features! LGTM!

building_sim_plugins/building_gazebo_plugins/src/crowd_simulator.cpp
aaronchongth: There is not that clear a solution I believe. It will ultimately depends on how you would like the plugin's behaviour to be, and how often would you expect these edge cases to show up. If it is something that is critical for the continual operation of the plugin, you might want to crash the plugin while providing error logs. Otherwise if this happens once in a while, as a result of dropping a sensor message for example, and you would like the plugin to continue, then an error log while calling `return;` on the function would be good.

building_sim_plugins/building_ignition_plugins/CMakeLists.txt
```diff
@@ -130,6 +131,43 @@ target_link_libraries(lift
...
+  )
+
+  target_link_libraries(crowd_simulator_ign
+    PUBLIC
+    ignition-gazebo${IGN_GAZEBO_VER}::ignition-gazebo${IGN_GAZEBO_VER}
+    ignition-plugin${IGN_PLUGIN_VER}::ignition-plugin${IGN_PLUGIN_VER}
+    ${rclcpp_LIBRARIES}
+    ${building_sim_common_LIBRARIES}
+    ${menge_LIBRARIES}
+  )
```
luca-della-vedova, line 159: We should probably start moving the ignition plugins to `ament_target_dependencies`, the same as the other libraries.

FloodShao: Changed that part in CMakelists

luca-della-vedova:
Great thanks for all the changes!

luca-della-vedova (approved):

PR #220, opened 2020-08-20
Adding lift operation range selection

MakinoharaShouko:
As discussed in [a previous pull request](https://github.com/osrf/traffic_editor/pull/216), there should be some way to select the floors on which the lift can reach so that the traffic generator does not generate unwanted holes. 

This pull request adds the dropdown menus to select the highest and the lowest floors the lifts can reach. Accordingly, the generation process will choose to generate holes only on levels in between.

The problem of lift falling appeared after adding this feature is caused by the lift plugin and is being fixed in [this branch](https://github.com/osrf/traffic_editor/tree/fix/lift_initial_floor)

The pull request also contains some other minor fixes, such as commenting out some unnecessary lines from `LiftDialog::ok_button_clicked` since these values are already assigned when the user edits them and there is no need to do the assigning one more time when clicking the ok button


codebot (approved):

MakinoharaShouko:
> Thanks for all of this work! 🎉 Looks good to me, let's get this in. Based on the comment above:
> 
> > The problem of lift falling appeared after adding this feature is caused by the lift plugin and is being fixed in this branch
> 
> Does this need a simultaneous merge with another PR?

@Kevinskwk is working on this and his PR would be ready after this one is merged

traffic_editor/gui/lift.cpp
```diff
@@ -193,11 +211,22 @@ void Lift::draw(
...
 bool Lift::level_door_opens(
   const std::string& level_name,
-  const std::string& door_name) const
+  const std::string& door_name,
+  const std::vector<BuildingLevel>& levels) const
 {
   LevelDoorMap::const_iterator level_it = level_doors.find(level_name);
   if (level_it == level_doors.end())
     return false;
+  for (auto level : levels)
```
Yadunund, line 220: Pedantic suggestion
```suggestion
  for (const auto& level : levels)
```

traffic_editor/gui/lift.cpp
```diff
@@ -52,6 +53,18 @@ void Lift::from_yaml(const std::string& _name, const YAML::Node& data)
     }
   }
 
+  if (data["highest_floor"])
+    highest_floor = data["highest_floor"].as<string>();
+  if (data["lowest_floor"])
+    lowest_floor = data["lowest_floor"].as<string>();
+  for (auto level : levels)
```
Yadunund, line 60:
```suggestion
  for (const auto& level : levels)
```

traffic_editor/gui/lift_dialog.cpp
```diff
@@ -76,6 +77,70 @@ LiftDialog::LiftDialog(Lift& lift, Building& building)
...
+    _highest_floor_combo_box,
+    &QComboBox::currentTextChanged,
+    [this](const QString& text)
+    {
+      _lift.highest_floor = text.toStdString();
+      if (_lift.highest_floor.empty())
+        _lift.highest_elevation = DBL_MAX;
+      else
+      {
+        for (auto level : _building.levels)
```
Yadunund, line 98:
```suggestion
        for (const auto& level : _building.levels)
```

MakinoharaShouko:
Yeah this one can be merged

traffic_editor/gui/lift_dialog.cpp
```diff
@@ -76,6 +77,70 @@ LiftDialog::LiftDialog(Lift& lift, Building& building)
...
+    _lowest_floor_combo_box,
+    &QComboBox::currentTextChanged,
+    [this](const QString& text)
+    {
+      _lift.lowest_floor = text.toStdString();
+      if (_lift.lowest_floor.empty())
+        _lift.lowest_elevation = -DBL_MAX;
+      else
+      {
+        for (auto level : _building.levels)
```
Yadunund, line 130:
```suggestion
        for (const auto& level : _building.levels)
```

Yadunund (changes requested):
Thanks for adding this feature! Its looking great. Left a couple of minor pedantic suggestions before we can merge this.

MakinoharaShouko:
OK I have just committed the suggestions

Yadunund (approved):

PR #221, opened 2020-08-21
Improve lift initial floor definition

Kevinskwk:
This PR includes the following features:
- Adding a field in the lift dialog of traffic_editor GUI to specify the initial floor of the lift
- When the initial floor is not being served by the lift, the lift will be generated on the first floor instead
- If a lift is not serving any floor (i.e. having no doors open on any floor), the lift model will not be generated.

I am making this a draft PR, for now, to avoid conflicts with https://github.com/osrf/traffic_editor/pull/220. Once that is merged, I might change the lift shaft holes generation when lift model is not generated accordingly.

### DIFF
--- a/building_map_tools/building_map/building.py
+++ b/building_map_tools/building_map/building.py
@@ -216,6 +216,9 @@ class Building:
             pose_ele.text = f'0 0 {level.elevation} 0 0 0'
 
         for lift_name, lift in self.lifts.items():
+            if not lift.level_doors:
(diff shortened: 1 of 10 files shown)


codebot (approved):
`traffic-editor` compiles and runs nicely for me with this PR on the multi-level `clinic` world. @Yadunund do you have any specific feedback? Otherwise, looks good to me :+1:

Yadunund:
Looks good to me!

PR #222, opened 2020-08-25
Fred/slotcar fixing

FloodShao:
Currently the slotcar plugin directly uses the vector distance to check whether there is an emergency stop to avoid the collision.
This distance should be checked in 2D, because some models position has an elevation in z direction especially some actor model. 
e.g. currently the stop_distance_threshold is set at 0.75, but the model position is at (x, t, 1.0), wherever the model at, the distance checked is larger than stop_distance_threshold, as such, no emergency stop will be triggered.


codebot (approved):

codebot (changes requested):
Hmm this is a tricky one, because we don't want to see models on different building levels trigger an e-stop of slotcars (for example, a robot on L5 and a robot on L2 shouldn't interfere). But yeah if one is at +0.1m above the floor and another is +0.0 , they should interfere. Maybe there should be a threshold on this, or a check (if available) of the level names? Not sure of the cleanest approach.

FloodShao:
Is there a threshold for level elevation? How about checking the z difference? This will be checked only if the z difference of two models is smaller than the adjacent level elevation.

FloodShao:
The `obstacle_positions` only excludes all the infrastructure (lifts, and doors). It doesn't check for current level.

FloodShao:
Just added.

building_sim_plugins/building_plugins_common/src/slotcar_common.cpp
```diff
@@ -296,7 +296,11 @@ bool SlotcarCommon::emergency_stop(
   bool need_to_stop = false;
   for (const auto& obstacle_pos : obstacle_positions)
   {
-    if ((obstacle_pos - stop_zone).norm() < _stop_radius)
+    Eigen::Vector3d detect_vector = obstacle_pos - stop_zone;
+    // check whether the obstacle model is in the same level with the slotcar, hardcode adjacent level elevation 2.0 m
+    if (std::abs(detect_vector(2)) > 2.0) continue;
```
luca-della-vedova, line 301: Is the level elevation in traffic editor a fixed parameter? I wonder if there is a way to make this value a variable that is fetched from the world settings?

codebot: I think the use of `2.0` here is for the vertical difference between the two objects, not the absolute elevation.

luca-della-vedova: Yea but if level differences can be changed this could still lead to unpredictable behavior. For example if there was a building with very low levels (let's say 1.5m) the slotcar would stop for objects that are upstairs, if the levels were very high and there was a very tall object that is over 2 meters (i.e. a tall pillar in a warehouse) the slotcar would just go through it.

codebot: i believe that our current convention is that the Z height offset of a model is the "base" of the model. So for example, pillars and people and robots and everything should have Z=0. This exception here is (so far as I understand it) for things like coffee mugs on tabletops or things like that. Fred, is that how you see it also? Or do we have some models that have nonzero Z-offset values and need to be translated up/down so that they look "correct" at z=0 ?

FloodShao: This issue occurs because the common actors I used in crowd simulation has a z=1.0 (on the waist of the human body), and currently the slot car is using 0.75 as the stop_range, so there is no way the e-stop will be triggered. I don't know if there are already designed human models can be used (with base position z = 0). If there are, I can try for that. 
Now the tall pillar situation:
the `stop_zone` point is at the same level with the slot car (z_slotcar = z_level_elevation)
the `obstacle_pos` is the designed pose of the object (x_pillar, y_pillar, z_pillar). 
* I think we should talk about the pillar will not go through all levels right? like one pillar should remain within one level, this pillar should not pass through several levels. 
if z_pillar = z_level_elevation, then, z_pillar = z_slotcar, definitely will trigger the e-stop.
if z_pillar is in the middle of the pillar model, then z_level_elevation < z_pillar < z_level_elevation + 2.0, this will be captured by my fixed code.
* If the pillar pass through multiple levels, then the e-stop trigger might be random at different level.

codebot: Oh ok, sorry, I had misunderstood this issue totally. Yeah, I think the best fix for this issue is actually to make sure that all the models we are using have z=0 in order to look reasonable when "on the ground." I would suggest looking at the new hospital humans and seeing if they look reasonable at z=0. If not, we can definitely modify them.

As for the pillars, currently they are generated as part of the walls of each floor. There are no multi-floor pillars, to my knowledge. They're just wall segments, arranged in a box shape. We do have some circular pillar models for car parks and loading bays, but they are just "normal" static models that should render correctly as z=0.

FloodShao: OK, sure.

PR #223, opened 2020-08-28
Update lift session id in lift plugin

Kevinskwk:
This PR updates lift `session_id` in the `lift_state` message each lift plugin stores and publishes. This is required for the new features in `rmf_core` to make robots check the `session_id` before entering the lift.

Additionally, a minor bug in `available_floors` displayed in `lift_state` messages is fixed.


Yadunund (approved):
LGTM!

PR #224, opened 2020-08-31
Fred/building_crowdsim for generating the navmesh file and required configuration files for menge

FloodShao:
This sub-package is developed for generating the navmesh file and required configuration files from the building.yaml output by traffic-editor GUI.
This package is integrated within the `building_map_tools` package, and can be called using ros2 run command.
Before using `building_crowdsim`, the building.yaml files with "human_lanes" and "crowd_sim" tag. An example building.yaml for office scenario can be found here: https://github.com/FloodShao/traffic_editor/tree/fred/building_crowdsim_test/building_crowdsim_test_files

ros2 run command, after package building:
1. for generating the navmesh using "human_lanes"
`ros2 run building_map_tools building_navmesh navmesh [input_building_yaml] [output_dir]`
2. for generating the configfiles using "crowd_sim"
`ros2 run building_map_tools building_navmesh configfile [input_building_yaml] [output_dir] [world.sdf_to_be_inserted_with_plugin] [gazebo_or_ign_tag]`


luca-della-vedova:
I see a bit of test code scattered around the code base, specifically in

- building_navmesh.py
- navmesh/test.py

 I guess it could either be moved to actual unit / integration test files or deleted, what do you think?

luca-della-vedova:
Another "high level" comment is that I would try to follow the conventions in the rest of the code base, from a quick glance (similar to the plugins PR):

- snake_case for function names and general style consistency (I see some snake case, some camelCase, some UpperCamelCase for function names scattered around). IIRC we generally follow [PEP8](https://www.python.org/dev/peps/pep-0008/) for Python.
- getters / setters are [not typically used in Python](https://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters) and attributes are usually accessed directly, this should reduce the code size by quite a lot in simpler files, such as most of the smaller files under `navmesh`

building_map_tools/building_crowdsim/building_yaml_parse.py
```diff
@@ -0,0 +1,89 @@
...
+        self.human_goals = []
+
+        self.calculate_scale_using_measurements()
+        self.transform_all_vertices()
+        self.update_human_goals()
+
+    def get_transformed_vertices(self):
+        return self.transformed_vertices
+
+    def get_human_lanes(self):
```
aaronchongth, line 33: What Luca mentioned, we normally avoid getters and setters in python codebase

building_map_tools/building_crowdsim/building_yaml_parse.py
```diff
@@ -0,0 +1,89 @@
...
+            self.levels_name.append(level_name)
+
+        # crowd_sim for configuration
+        if not 'crowd_sim' in self.yaml_node:
+            # TODO: add exit process
+            print("Expected 'crowd_sim' tag for crowd simulation")
+            
+        self.crowd_sim_config = self.yaml_node['crowd_sim']
+
+    def GetLevelNames(self):
```
aaronchongth, line 75: There seems to be a style difference, using UpperCamelCase vs snake_case, do consider using the same style check mentioned in the previous PR, https://github.com/osrf/rmf_core/blob/feature/style_action/.github/workflows/style.yaml to check locally

building_map_tools/building_crowdsim/navmesh/vector.py
```diff
@@ -0,0 +1,37 @@
...
+    def getDot(self, vector2d):
+        return self.x * vector2d.x + self.y * vector2d.y
+
+    def getOrientation(self):
+        return math.atan2(self.y, self.x)
+    
+    def getCross(self, vector2d):
+        return self.x * vector2d.y - self.y * vector2d.x
+
+if __name__ == "__main__":
```
aaronchongth, line 34: Looks like some testing code is still here

building_map_tools/building_crowdsim/navmesh/build_navmesh.py
```diff
@@ -0,0 +1,106 @@
...
+    buildNavmesh.AddLane(8, 9, 1.5)
+    buildNavmesh.AddLane(7, 8, 1.5)
+    buildNavmesh.AddLane(0, 7, 1.5)
+    buildNavmesh.AddLane(7, 10, 1.0)
+
+    buildNavmesh.Process()
+    buildNavmesh.Output("test_navmesh_00.nav")
+
+if __name__ == "__main__":
+    test()
```
aaronchongth, line 106: Test found here, could be put into a separate unit test

building_map_tools/building_crowdsim/config/template_conf_yaml.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def load(self, external_agent_name_list):
+        for name in external_agent_name_list :
+            self._external_agent.add(str(name))
+        
+    def getExternalAgents(self) :
+        return list(self._external_agent)
+
+
+if __name__ == '__main__' :
```
aaronchongth, line 291: tests found here

aaronchongth (changes requested):
Thanks for putting these features together, @FloodShao!

I am hesitant to review this fully at the moment, as it requires a lot of workarounds to test it with the other branches, namely, https://github.com/osrf/traffic_editor/pull/218 and https://github.com/osrf/traffic_editor/pull/225. I will suggest we merge https://github.com/osrf/traffic_editor/pull/218 first, rebase this branch, properly integrate the changes so that we can test the whole build pipeline.

@luca-della-vedova and I discussed a bit regarding maintainability, and would like your opinion on this. Since we are not writing APIs, but rather declaring a lot of different data structures, do you think it will be a better idea to keep the code base small by accessing and manipulating the data structures via their public members, instead of getters and setters?

Only when further conditions are needed to access data from data structures (eg. checks, calculations, etc), and that it is widely used across the code base, then we should consider defining getters for that.

The source files are currently missing their respective licenses, and the style should also be checked.

FloodShao:
Agreed. We can merge #218 first, and I'm working on #225 and #224. Thanks for all that comments and advice on the code, that helps a lot. 
Yes, I will try to avoid the getters and setters in python class (I should have checked with @luca-della-vedova first). and that definitely will make the code small.
I will modify the code style and licenses as well. Thanks!

FloodShao:
Hi, @aaronchongth and @luca-della-vedova, I have reworked the class structure, removed unnecessary setter and getter functions, and modified the code style. I have rebased this pr last THU, so this PR includes the crowd_simulation_plugin now.
As for the test, I created a test branch base on this pr, in which a CMake base package in added to run the pipeline. <https://github.com/FloodShao/traffic_editor/tree/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test>.
Now you can follow the steps to test it:
1. use followsing repose:
```
repositories:
  rmf/rmf_core:
    type: git
    url: https://github.com/osrf/rmf_core.git
    version: master
  rmf/rmf_schedule_visualizer:
    type: git
    url: https://github.com/osrf/rmf_schedule_visualizer.git
    version: master
  rmf/traffic_editor:
    type: git
    url: https://github.com/FloodShao/traffic_editor.git
    version: fred/building_crowdsim_navmesh_test
  rmf/rmf_demos:
    type: git
    url: https://github.com/osrf/rmf_demos.git
    version: master
  meng_core:
    type: git
    url: https://github.com/FloodShao/menge_core.git
    version: master
```
2. colcon build the whole rmf workspace.
3. find the launch file in source folder (because I didn't install the launch file in the package share folder). 
for office scenario: 
<https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/maps/office/office.crowd_simulation.launch.xml>
for airport scenario:
<https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/maps/airport_terminal/airport_terminal.launch.xml>
4. launch the command:
for office scenario: `ros2 launch office.crowd_simulation.launch.xml`
for airport_scenario: `ros2 launch airport_terminal.launch.xml gazebo_version:=9` if you only have gazebo9 now.

I think these steps should be able to test the pipeline after getting the "building.yaml". 

Thank you for your time and patient for reviewing this PR.

luca-della-vedova:
I'm looking at the generation [CMakeLists](https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/CMakeLists.txt) and trying to understand the steps and if they can be simplified or documented better.

From what I understand there is need to add three build items to the CMakeLists if anyone was to add crowd simulation to their world:

1. The [navmesh generation](https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/CMakeLists.txt#L60), which is the format that Menge expects to move the actors around.
2. The [configfile generation](https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/CMakeLists.txt#L69), Another menge specific set of xml files (behavior_file and scene_file that define respectively the behavior of agents and their number / category / spawning points).
3. The [nav graphs generation](https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/CMakeLists.txt#L83) that I'm guessing is used by the simulation plugin to control the actors (?).

I wonder if there is any merging that can be done for those, I think the only reason we would want them separate would be if there is a use case for generating only a subset of them and it seems that anyone who wants to integrate crowd simulation will always need to generate all the three sets of files?
The idea behind the merging would be to make it easier for people to integrate, right now it seems they would need to add three custom commands with the right parameters or it would just not work.

Also I don't really understand the [phony](https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/CMakeLists.txt#L83) folder, what is its purpose?

FloodShao:
Hi @luca-della-vedova , only <https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh_test/building_crowdsim_pr_test/CMakeLists.txt#L44-#L75> is related to the crowd simulation setup. Only these block is inserted compared to the <https://github.com/osrf/rmf_demos/blob/master/rmf_demo_maps/CMakeLists.txt>
step 1 and step2 you are right. the nav_garphs generation is for the robot delivery rviz. 

Yep, I think I can merge the two commands together into one if that's necessary. Because step 1 and step 2 are processing the same building.yaml, except step 2 requires a generated world.sdf as input.

building_map_tools/building_crowdsim/navmesh/object_manager.py
```diff
@@ -0,0 +1,9 @@
+class Manager:
+
+    def __init__(self):
+        self.data = []
+
+    def add_obj(self, obj):
+        assert(hasattr(obj, 'id')), "Adding object without id attribute!"
+        obj.id = len(self.data)
+        self.data.append(obj)
```
luca-della-vedova, line 9: I'm not too sure what's the idea behind this object, is it necessary?

building_map_tools/building_crowdsim/navmesh/polygon.py
```diff
@@ -0,0 +1,76 @@
...
+        self.calulate_center()
+        result.append(self.center.get_coords())
+        vertices_id = []
+        for v in self.vertices:
+            vertices_id.append(v.id)
+        result.append(vertices_id)
+        result.append(self.gradient)
+        result.append(self.edge_ids)
+        result.append(self.obstacle_ids)
+        return result
```
luca-della-vedova, line 46: I'm generally concerned with the `get_variable` functions you introduce in your classes. Imho it would be a lot clearer if downstream users just received the object and accessed the variables they need by name (i.e. `poly.gradient` or `poly.edge_ids`) instead of packing everything into a list and having to access it by index (that at the end is kind of a magic number)

building_map_tools/building_crowdsim/navmesh/polygon_factory.py
```diff
@@ -0,0 +1,397 @@
...
+        self.lane_vertex_manager = self.connection_manager.lane_vertex_manager
+        self.lane_manager = self.connection_manager.lane_manager
+        self.polygon_vertex_manager = polygon_vertex_manager
+        self.edge_manager = edge_manager
+        self.obstacle_manager = obstacle_manager
+        self.polygon_manager = polygon_manager
+        # predefined scale factor, see self.intersectionPolygonUpdate() special
+        # case explaination
+        self.scale_special_case = 0.01
+        self.dead_end_extension = 0.1
```
luca-della-vedova, line 31: Can you add a comment on what these values represent?

building_map_tools/building_crowdsim/navmesh/polygon_factory.py
```diff
@@ -0,0 +1,397 @@
...
+
+        v2 = construct_vertices[0]
+        v3 = construct_vertices[1]
+        self.polygon_vertex_manager.add_obj(v2)
+        self.polygon_vertex_manager.add_obj(v3)
+
+        if self.connection_manager.is_on_same_side_of_lane(lane.id, v0, v2):
+            result = [v0, v2, v3, v1]
+        else:
+            result = [v0, v1, v2, v3]
```
luca-della-vedova, line 266: There is a similar rearrangement logic in `lane_polygon_general_case`, can the logic be moved to a shared function?

building_map_tools/building_crowdsim/navmesh/polygon_factory.py
```diff
@@ -0,0 +1,397 @@
...
+        # dead end case, one more obstacle added
+        if len(lane_polygon.edge_ids) == 1:
+            edge_id = list(lane_polygon.edge_ids)[0]
+            edge = self.edge_manager.data[edge_id]
+            on_edge_vertex_ids = [edge.v0_id, edge.v1_id]
+
+            obstacle_pairs = []
+            for v in polygon_vertices:
+                if v.id not in on_edge_vertex_ids:
+                    obstacle_pairs.append(v.id)
```
luca-della-vedova, line 389: I think you can make this a one liner by using list comprehension

building_map_tools/building_crowdsim/navmesh/obstacle.py
```diff
@@ -0,0 +1,17 @@
...
+        self.neighbor_obst = -1
+
+    def init_obstacle(self, v0_id, v1_id, node_id, neighbor_obst=-1):
+        self.v0_id = v0_id
+        self.v1_id = v1_id
+        self.node_id = node_id
+        self.neighbor_obst = neighbor_obst
+
+    def get_variable(self):
+        return [self.v0_id, self.v1_id, self.node_id, self.neighbor_obst]
```
luca-della-vedova, line 17: Same comment about usage of get_variable when you can just access by name

building_map_tools/building_crowdsim/test/build_configfile_TEST.py
```diff
@@ -0,0 +1,131 @@
...
+from building_crowdsim.config.scene_file import\
+    SceneFile, ObstacleSet, AgentProfile, AgentGroup
+from building_crowdsim.config.behavior_file import\
+    BehaviorFile, BehaviorState, StateTransition, GoalSet
+from building_crowdsim.config.plugin_file import Plugin
+import building_crowdsim.config.util as util
+from building_crowdsim.building_yaml_parse import BuildingYamlParse
+
+
+def simple_unit_test():
```
luca-della-vedova, line 14: First of all I would rename the test to be a bit more explicative, also I don't quite understand what the tests are supposed to check since there is no checking of the generated files. Do the only test that the functions managed to run without triggering one of the raise / asserts?

building_map_tools/building_crowdsim/test/build_navmesh_Base_TEST.py
```diff
@@ -0,0 +1,24 @@
+from building_crowdsim.navmesh.build_navmesh import BuildNavmesh
+
+
+def test():
```
luca-della-vedova, line 4: Same as the comment for the `build_configfile_TEST.py` I also wonder whether those tests are integrated with colcon so that they are ran when doing `colcon test`?

luca-della-vedova (changes requested):
OK! Had some time to go through the code and give some more detailed feedback, for the CMakeLists I think we should try to integrate them in a single command so there is less risk for the downstream users to make mistakes (if for example they only call one of the two commands).

building_map_tools/building_crowdsim/config/behavior_file.py
```diff
@@ -0,0 +1,287 @@
...
+        self.attributes['name'] = ''
+        self.attributes['final'] = 0
+        self.goal_selector = GoalSelector()
+        self.vel_component = VelComponent()
+
+    def is_valid(self):
+        if not self.attributes['name']:
+            return False
+        if self.attributes['final'] == 1:
+            return True
```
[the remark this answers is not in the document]
FloodShao: There are 2 types of states. Both types require "name" attributes. If the state is final, we don't need to check other attributes as other attributes will not be provided. If the state is not final, we will have to check pther attributes. The return can be removed but the other attributes check should be put inside the if condition...

building_map_tools/building_crowdsim/config/plugin_file.py
```diff
@@ -0,0 +1,175 @@
...
+from .leaf_element import LeafElement, Element
+
+
+class Plugin (Element):
+    def __init__(self, model_env):
+        Element.__init__(self, 'plugin')
+        if model_env == "gazebo":
+            self.attributes['filename'] = 'libcrowd_simulator.so'
+        elif model_env == 'ign':
+            self.attributes['filename'] = 'libcrowd_simulator_ign.so'
```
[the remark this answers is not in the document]
FloodShao: Oh, then I will have to change the CMakeLists in the building_ignition_plugins first, because the plugin name is specified there.

building_map_tools/building_crowdsim/config/plugin_file.py
```diff
@@ -0,0 +1,175 @@
...
+        self.attributes['name'] = 'crowd_simulation'
+
+        self.behavior_file_element = LeafElement('behavior_file')
+        self.behavior_file_element.text = 'behavior_file.xml'
+
+        self.scene_file_element = LeafElement('scene_file')
+        self.scene_file_element.text = 'scene_file.xml'
+
+        self.update_time_step_element = LeafElement('update_time_step')
+        self.update_time_step_element.text = '0.1'
```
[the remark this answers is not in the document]
FloodShao: This is exposed in the GUI. This 2 lines are in the class constructure. "update_time_step" will be set to user defined value (default is 0.1) in load_from_yaml, check
https://github.com/FloodShao/traffic_editor/blob/fred/building_crowdsim_navmesh/building_map_tools/building_crowdsim/config/plugin_file.py#L37-#L40

building_map_tools/building_crowdsim/config/plugin_file.py
```diff
@@ -0,0 +1,175 @@
...
+        Element.__init__(self, 'model_type')
+        self.type_name = LeafElement('typename')
+        self.animation_speed = LeafElement('animation_speed')
+        self.animation = LeafElement('animation')
+        # for gazebo model
+        self.initial_pose_gazebo = LeafElement('initial_pose')
+        self.gazebo_filename = LeafElement('filename')
+        # for ign model
+        self.initial_pose_ign = LeafElement('initial_pose')
+        self.ign_filename = LeafElement('filename')
```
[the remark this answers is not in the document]
FloodShao: This is a bit tricky. Because currently, to spawn an actor, gazebo requires to provide .dae file like
`model://MaleVisitorPhone/meshes/MaleVisitorPhoneWalk.dae`. If the idle animation is provided, another idle.dae should be provided. Ignition only needs a http dir like `https://fuel.ignitionrobotics.org/1.0/OpenRobotics/models/Male%20visitor` or `model://MaleVisitorPhone` . The initial pose can be suppressed though if all the models exsits locally. What's your opinion on this?

building_map_tools/building_crowdsim/navmesh/navmesh_generator.py
```diff
@@ -0,0 +1,106 @@
...
+            level_with_human_lanes,
+            level_name,
+            default_graph_idx=9):
+        assert(isinstance(level_with_human_lanes, LevelWithHumanLanes))
+        self.level = level_with_human_lanes
+        self.level_name = level_name
+        self.current_graph_idx = default_graph_idx
+
+    def set_graph_idx(self, graph_idx):
+        self.current_graph_idx = graph_idx
```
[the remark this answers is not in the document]
FloodShao: This setter is never used. So I forgot to delete it... I will just delete it.

building_map_tools/building_crowdsim/navmesh/object_manager.py
FloodShao: This is the base class for all the VerticesManager, LaneManager, PolygonManager... The idea is all these manager will initialize the id to each added instance.

building_map_tools/building_crowdsim/navmesh/polygon.py
FloodShao: Oh, this get_variable is called when writing the actual navmesh to the file. When writing the navmesh, the manager will go over all the vertices, polygons, edges, obstacles, and write the corresponding params in it.
It's like the building.yaml. All the lanes, edges, models are construct based on the vertices. Each vertex has one unique id, and each lane only store the related vertices id rather than store the actual vertices instance. 
Similarly, polygon only store the ids of related vertices, ids of related edges, ids of related obstacles. And each edge only store the related vertices ids, the same as obstacles.
That's why the above "object_manager" is needed to assign a unique id for each added instance.

building_map_tools/building_crowdsim/navmesh/obstacle.py
FloodShao: It's the same purpose as the polygon.get_variable. The purpose if not provide a variable access. The function is used for formatting the navmesh output.

building_map_tools/building_crowdsim/test/build_configfile_TEST.py
FloodShao: For the configfile generation, I guess I put lots of raise ValueErro in the code, so that if the test function successfully ends without any ValueError, the test should be fine. When dealing with Navmesh, even though the function ends without any assertion and ValueError, the navmeh itself might be inappropriate in menge's logic due to some parameter selection. The base_test provides a simple test case full of special cases so that we can check whether the generated file is correct. As for the configfile, it actually just transforms the yaml configuration to the xml format under the menge definition. With the raise and is_valid function defined in each type, I think if the test function successfully ends the test should be fine. Although I don't know whether put checks inside the code is the proper way to do it.

building_map_tools/building_crowdsim/config/behavior_file.py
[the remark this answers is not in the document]
luca-della-vedova: Oh I see, makes sense

FloodShao:
Current PR changes:
build and tested under focal/foxy/gazebo11
"plugin" part:
1. Modify the ign CMakeLists to change the ign plugin name the same as gazebo
2. Modify the gazebo actor spawn method to to make sure the plugin parameter the same as ign. **Now the simulation only support gazebo11!**
"python" part:
1. Make the ModelType class platform independent.
2. Use one line command `ros2 run building_map_tools building_crowdsim [.yaml] [output_dir] [world.sdf]` to generate all the required files for menge
3. Add tests to package setup. compare the navmesh test with the standard result. Now the tests can be conducted under `colcon test` and checking the test result under `colcon test-result`
(Thanks to Brandon's <https://github.com/colcon/colcon-ros/issues/105>)

Test:
In branch <https://github.com/FloodShao/traffic_editor/tree/fred/building_crowdsim_pr_test>.
Both gazebo and ign launch file for office scenario are provided. Airport scenario only provided with gazebo launch file.
Environment: 
1. Must be under gazebo 11! Although I think eloquent + gazebo11 should also work.
2. Please make sure you have `model://MaleVisitorPhone` in your `~/.gazebo/models` (cannot directly use the walk.dae anymore after changes the actor spawn method)
Steps:
1. Use the above branch for "traffic_editor"
2. `colcon build`
3. Go to `src/rmf/traffic_editor/building_crowdsim_pr_test/maps/office`.
use `ros2 launch office.launch.xml` or `ros2 launch office_ign.launch.xml` for office scenario

luca-della-vedova:
Running colcon test I found a test failure in the package:

```
</system-out></testcase><testcase classname="building_map_tools.test.building_crowdsim.test_build_configfile" file="test/building_crowdsim/test_build_configfile.py" line="83" name="test_plugin_from_yaml" time="0.181"><failure message="TypeError: __init__() takes 1 positional argument but 2 were given">test/building_crowdsim/test_build_configfile.py:89: in test_plugin_from_yaml
    plugin_file = Plugin(&apos;gazebo&apos;)
```

FloodShao:
> Running colcon test I found a test failure in the package:
> 
> ```
> </system-out></testcase><testcase classname="building_map_tools.test.building_crowdsim.test_build_configfile" file="test/building_crowdsim/test_build_configfile.py" line="83" name="test_plugin_from_yaml" time="0.181"><failure message="TypeError: __init__() takes 1 positional argument but 2 were given">test/building_crowdsim/test_build_configfile.py:89: in test_plugin_from_yaml
>     plugin_file = Plugin(&apos;gazebo&apos;)
> ```

Apologies for that. This was caused by the changing Plugin constructor. The previous Plugin constructor requires one parameter for the platform. And I forgot to changed that line in the test file.

FloodShao:
Just fixed the deleting file problem and the test failure problem (it turns out the I forgot to update the test yaml file to be consistent with the current code as well.) 
Now I think nothing is broken, and all tests pass.

luca-della-vedova (approved):
Awesome, thanks for all the changes!

PR #225, opened 2020-08-31
Fred/crowdsim_gui modifications on traffic_editor for configuring crowd simulation

FloodShao:
This PR includes the modifications on traffic_editor gui for the purpose of generating the "human_lanes" and "crowd_sim" tag in building.yaml. The building.yaml will then be processed with 'building_map_tools:building_crowdsim' to generate the navmesh and required config files for menge.

This integration can be used once passing the package building. A documentation on using the crowd_sim panel will be written, and a pop-up widget might be integrated in this. Currently a structure review may be helpful before I start adding more code to it.

An example crowd_sim configuration file for office scenario can be found in here:
<https://github.com/FloodShao/traffic_editor/tree/fred/building_crowdsim_test/building_crowdsim_test_files/map>


codebot (changes requested):
A few review comments are scattered throughout. A larger issue is with `crowd_sim_dialog.h` and  `crowd_sim_dialog.cpp` which contain many different classes. I know it's convenient to keep these subclasses together, but it does get somewhat confusing when paging through the file.

I'd suggest to add a subdirectory called `gui/croud_sim` and break apart `crowd_sim_dialog.h` and `crowd_sim_dialog.cpp` into the usual one-class-per-file, in that directory.

FloodShao:
Hi @codebot , thanks for the comments on this PR. 
In this PR, the major logic is to create a table to show the configuration, and put that table into a dialog with ok and cancel button widget. 
I have made changes on this PR, the major changes includes:
1. separate all the required class headers, group them in "include/crowd_sim/"
2. separate all the tables into one class one file format, group them in "gui/crowd_sim"
3. create a pure virtual class for all the tables to reduce the code size
4. as for all the dialogs, the class declare and definition are quite similar (except for condition_dialog.h). I grouped them into "crowd_sim_dialog.h" and "crowd_sim_dialog.cpp" to avoid generating multiple headers and cpp. All the .cpp for dialogs (except for condition_dialog.cpp) only includes the constructor. I'm not sure if that's appropriate. If not, I can separate all of them.
5. for member functions only have less than 3 lines, I write them in the header file as inline function. Mostly are setters and getters. If that's not appropriate, I can modify them as well.

Thanks for your time and patience reviewing this PR.

FloodShao:
Hi @codebot , I managed to make all the classes have their own header file and implementation. I reduced some redundant code and use a base class pointer for tables to be shown in the dialog, so that we won't have to create all the dialog classes.
Besides, I added all the license declaration into the front of each source file. 
If you have further comments please let me know. Thanks for reviewing this PR.

Issue #219, opened 2020-08-20
Deselecting a floor in Gazebo does not hide the models on the floor


mxgrey:
As shown in the screenshot below, when you toggle off the visibility for a floor, the models on that floor remain visible. It would probably be better if those could also be turned off, so we don't see mysterious floating beings.

![world_gzclient_camera(1)-2020-08-20T17_00_41 414442](https://user-images.githubusercontent.com/1307217/90750077-e46ed880-e306-11ea-89db-a84e637c9d1a.jpg)

codebot:
Ah! Indeed. That's not super great. Thank you for finding this.

codebot:
There are many ways to go about this, but at the moment it seems the simplest way would be to look at the Z height of the floor that is being toggled, and then use a heuristic window around that Z height (say, -0.2 through +2.5 meters) and toggle all models (static only?) that lie within that Z window. Sound OK?

mxgrey:
That sounds okay for a first-order approach. I think there could be some loose ends if an object has a wonky origin (leading to false positives or false negatives), but that simple approach should cover 99% of cases.

codebot:
Cool. Then we can avoid (at least for now) adding some sort of metadata to the automatically-generated models and trying to fish it out in the plugin later. Using the Z window feels like a minimal first attempt, and we can see how it goes.

Yadunund:
@Kevinskwk has proposed a method where a list of static model names is appended as sdf subelements of the `toggle_floors` plugin in the `.world` file. https://github.com/osrf/traffic_editor/pull/226
When a level visibility is toggled, the plugin turns off the visibility of appropriate models from this list. On the pro side, it avoids having to manually define Z height boundaries but on the cons side, it increases the verbosity of the `.world` file. Any feedback on this approach?

mxgrey:
Sounds like a good approach. The last question would be whether we want to see robots or non-static models floating, and what should we do about it if we don't want that?

Yadunund:
We can extend the same behavior to non-static models such as doors. For robots things are trickier as they can move between levels. One option would be to subscribe to RobotState/FleetState messages and filter out robots whose reported levels match the toggle selection.

luca-della-vedova:
This was done some time ago in Gazebo and now also in Ignition, closing.

PR #226, opened 2020-08-31
Implement model visibility toggling

Kevinskwk:
This PR adds the feature of hiding & showing the static models on each level in the toggle_floors plugin. Static model names are recorded in the plugin sdf specification and hide/show messages are published on click of the toggle floor button.
Screenshot from 2020-08-31 17-56-53
Screenshot from 2020-08-31 17-57-01

### DIFF
--- a/building_sim_plugins/building_gazebo_plugins/src/toggle_floors.cpp
+++ b/building_sim_plugins/building_gazebo_plugins/src/toggle_floors.cpp
@@ -47,13 +47,12 @@ public:
         floor_ele->GetAttribute("model_name")->GetAsString();
 
       std::vector<string> models;
-      for (sdf::ElementPtr model_ele = floor_ele->GetFirstElement();
-        model_ele;
-        model_ele = model_ele->GetNextElement("model"))
+      auto model_ele = floor_ele->GetElement("model");
+      while (model_ele)
       {
-        if (model_ele->GetName() != string("model"))
-          continue;
-        models.push_back(model_ele->GetAttribute("name")->GetAsString());
+        if (model_ele->HasAttribute("name"))
+          models.push_back(model_ele->GetAttribute("name")->GetAsString());
+        model_ele = model_ele->GetNextElement("model");
       }
 
       printf(
--- a/building_sim_plugins/building_gazebo_plugins/src/toggle_floors.cpp
+++ b/building_sim_plugins/building_gazebo_plugins/src/toggle_floors.cpp
@@ -45,6 +45,17 @@ public:
       string floor_name = floor_ele->GetAttribute("name")->GetAsString();
       string model_name =
         floor_ele->GetAttribute("model_name")->GetAsString();
+
+      std::vector<string> models;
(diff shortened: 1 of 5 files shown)


building_sim_plugins/building_gazebo_plugins/src/toggle_floors.cpp
```diff
@@ -78,6 +90,11 @@ class ToggleFloors : public gazebo::GUIPlugin
     visual_msg.set_name(model_name);
     visual_msg.set_visible(visible);
     visual_pub->Publish(visual_msg);
+    for (string model : models)
```
Yadunund, line 93:
```suggestion
    for (const string& model : models)
```

building_sim_plugins/building_gazebo_plugins/src/toggle_floors.cpp
```diff
@@ -45,6 +45,17 @@ class ToggleFloors : public gazebo::GUIPlugin
       string floor_name = floor_ele->GetAttribute("name")->GetAsString();
       string model_name =
         floor_ele->GetAttribute("model_name")->GetAsString();
+
+      std::vector<string> models;
+      for (sdf::ElementPtr model_ele = floor_ele->GetFirstElement();
+        model_ele;
+        model_ele = model_ele->GetNextElement("model"))
+      {
+        if (model_ele->GetName() != string("model"))
```
Yadunund, line 54: I think we can get away with checking for this if we change the structure of the loop.
```
auto model_ele = floor_ele->GetElement("model")
while (model_ele)
{
  if (model_ele.HasAttribute("name")
    models.push_back(model_ele->GetAttribute("name")->GetAsString());
  model_ele = model_ele->GetNextElement("model");
}
```

Yadunund (changes requested):
Thanks for getting this out quickly! The approach is different from the one discussed in #219 but effective nonetheless. Lets reference this method and see what the best approach is. I've left a couple of minor comments for your consideration.

Yadunund:
Just a thought, should we also toggle visibility of doors on the level?

Kevinskwk:
I've implemented visibility toggling of doors (normal doors and lift shaft doors) on each level by appending the door model names in the same way. And I think it makes it look a lot better. While I don't think we should do the same to the non-static models. They are also hard to track since they might be moving between different levels.

Kevinskwk:
I've implemented the toggling of the robot models. The plugin stores the current floor visibilities and robot visibilities and checks if a robot's visibility has changed by checking its current floor in the `RobotState` msg. It's similar to what's done in the `fleet_state_visualizer` and it's slightly laggy due to the `RobotState` msg publishing interval.

I haven't come up with a method to track the current floors of other non-static models (e.g. coke cans in delivery). Any idea on this topic is appreciated.

tanyouliang95:
Tested this, and it works as expected! Thanks @Kevinskwk for the awesome work :tada:

Yadunund (approved):

PR #227, opened 2020-09-04
Create AxisAlignedBox component for slotcar

mrushyendra:
Create component in plugin's `Configure` method, to be used later for other purposes including dispensing onto/ingesting from slotcar.


luca-della-vedova (approved):
Looks good, looking forward to the delivery demos!

PR #230, opened 2020-09-16
Focal / Ignition dome dependencies update

luca-della-vedova:
Updated dependencies for ignition plugin after version bump in Dome.

Also fixed the traffic_editor package.xml ignition dependencies so they can be installed automatically through rosdep, keys have been added in focal for [ignition-plugin](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L1491-L1495) and [ignition-common3](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L1446-L1450).

Pondering adding a Focal / Foxy workflow, as well as building the plugins (for now the build is only running on the traffic_editor GUI), will probably do in a followup PR.


luca-della-vedova:
It shouldn't, the `package.xml` dependencies have never been used anyway and were just installed through apt-get in our workflows, the ignition dependency change just reflects the latest version bumps (and ignition builds are skipped unless ignition-dome is found, which doesn't happen in most setups since it has to be built from source).

Yadunund (approved):
LGTM

PR #233, opened 2020-09-18
fix #232 ensuring that building filename is set

### LINKED ISSUE #232: New projects have empty and useless building names and can't save
A regression has crept in so that when you create a new project from scratch (from an empty traffic-editor window) the default building filename is empty. This causes a silent error when you save and quit, and all your work is lost :skull:

codebot:
fixes #232 by adding a getter/setter for the `Project` class `filename` member, ensuring that whenever it's set, it validates that the building has a sane filename, and if not, it creates a default filename for it.

### DIFF
--- a/traffic_editor/gui/building.cpp
+++ b/traffic_editor/gui/building.cpp
@@ -388,8 +388,9 @@ void Building::set_model_yaw(
 
 void Building::clear()
 {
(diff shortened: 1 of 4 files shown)


Yadunund (approved):
Works great in my testing!

PR #236, opened 2020-09-25
Control slotcar with model velocity cmds in place of joint velocity cmds

mrushyendra:
Sends `LinearVelocityCmds` and `AngularVelocityCmds` to slotcars to mimic the motion that was earlier achieved using `JointVelocityCmds`. Ensures compatibility with both DART and the Trivial Physics Engine (TPE) plugin.


luca-della-vedova:
I gave it a quick spin and I found an issue that might be breaking in case of simulations with lifts.
You can see the issue below:

Slotcar falling velocitycmd

What I think is happening is that by setting VelocityCmd in DART we are basically bypassing the physics and hence other factors affecting the velocity (such as gravity in this case) are bypassed.
I don't expect us to have many cases of robots being in free fall but I'm a bit worried that if we had a robot on a lift and we set its speed to 0 it would bypass the gravity that makes it go down / collision force with the floor that make it go up and it would just hang in mid-air.

A hacky fix could be to set the Z `VelocityCmd` as the one from the current `Velocity`, but I can't tell on the top of my head if this would fix all the issues or not.

mrushyendra:
The `LinearVelocity` and `AngularVelocity` components aren't filled and remain at 0 always when using both DART and TPE. The `Pose` however seems to be updated, so it is possible to calculate the Z velocity using the change in pose and the delta time value. I tested this, and it _seems_ to work, but if the slotcar is ever to be manually moved by setting its pose value, that would throw the Z velocity calculation off.

luca-della-vedova:
There is generally two aspects to keep into consideration for the problem, the behavior in TPE and the behavior in DART.
From what I understand calculating the Z manually would fix the problem for DART (which is great) but we would still need to figure out how to get robots to move in lift with TPE since there would be no gravity. I'm conflicted between:

- Doing the "manual velocity calculation" you mentioned in DART, which from what I understand would solve the most general case (including robots in free fall, or even going up / down slopes) but would leave the TPE behavior to be figured out later.
- Finding a way to attach robots to lifts when they are moving, for example by setting their Z velocity to be equal to the lift's velocity. This would solve only the specific issue (lifts) but would work for both TPE and DART.

I am somewhat leaning towards the second, just to reduce the amount of work we need to do later for TPE, but I could be convinced otherwise.

What do you think?

mrushyendra:
It seems like the second option will be needed eventually for TPE anyway, so perhaps we could try implementing it first to see if it works with both engines. It seems like the plugin will need to detect whether it is on a lift, and if so, read the lift's velocity and/or position?

luca-della-vedova:
It might be better if the lift plugin itself detected if there is any models inside and set their velocity, the idea is that there might be other objects (i.e. a delivered object on top of the robot)

mrushyendra:
The slotcar plugin now monitors any payloads that gets dispensed to it, and moves the payloads as well if needed. In addition, the lift plugin searches for any objects that may have entered it, and sets their z-velocities to match the z-velocity of the lift, preventing the 'suspended in air' phenomena we saw earlier.

robot_lift-old
Old
robot_lift-new
New (Robot change is coincidental)

Also requires osrf/rmf_demos#150 to get merged.

Currently, it's not possible to test the lift behavior with the TPE plugin because doors and lifts have not been completely implemented yet. However, the fundamental method is the same as that used for delivery, or for lift travel with DART, so the behavior should be similar. Changing the step size from 0.001 to 0.01 may also help speed up testing with larger worlds for now.

Lastly, we also discussed whether it would be good to have the ability to control either joint velocities or model velocities based on the physics engine plugin used. The code for that would look like this (https://github.com/mrushyendra/traffic_editor/blob/slotcar_vel_cmds_with_joints/building_sim_plugins/building_ignition_plugins/src/slotcar.cpp). If that is preferred instead, I could merge those changes into this PR.

building_sim_plugins/building_ignition_plugins/src/slotcar.cpp
```diff
@@ -201,9 +239,53 @@ std::vector<Eigen::Vector3d> SlotcarPlugin::get_obstacle_positions(
   return obstacle_positions;
 }
 
+void SlotcarPlugin::charge_state_cb(const ignition::msgs::Selection& msg)
+{
+  dataPtr->charge_state_cb(msg.name(), msg.selected());
+}
```
luca-della-vedova, line 245: This change is related to the battery charging PR right? If there is too many of those changes might be worth it to move them to a separate PR, just to make tracking changes easier.

building_sim_plugins/building_ignition_plugins/src/slotcar.cpp
```diff
@@ -201,9 +239,53 @@ std::vector<Eigen::Vector3d> SlotcarPlugin::get_obstacle_positions(
...
+    this->_payloads.insert(new_payload);
+  }
+}
+
+void SlotcarPlugin::item_ingested_cb(const ignition::msgs::Entity& msg)
+{
+  if (msg.IsInitialized() && _payloads.find(msg.id()) != _payloads.end())
+  {
+    _payloads.erase(msg.id());
+  }
```
luca-della-vedova, line 262: Nit, we are doing two lookups in the `unordered_set` for the same `msg.id()`, I'd probably just create an iterator in the if statement, check it and erase it to do a single lookup.

building_sim_plugins/building_ignition_plugins/src/slotcar.cpp
```diff
@@ -201,9 +239,53 @@ std::vector<Eigen::Vector3d> SlotcarPlugin::get_obstacle_positions(
...
+    }
+    if (ecm.EntityHasComponentType(parent,
+      components::PhysicsEnginePlugin().TypeId()))
+    {
+      std::string physics_plugin_name =
+        ecm.Component<components::PhysicsEnginePlugin>(parent)->Data();
+      if (plugin_names.count(physics_plugin_name))
+      {
+        phys_plugin = plugin_names[physics_plugin_name];
+      }
```
luca-della-vedova, line 284: Same as above, can be replaced to do a single lookup through an iterator. Also I'd generally recommend against using the `operator[]` since it will do insertions when the key is not found (even though it's not an issue here). If you went for the iterator approach you should also be able to make the variable `const` I believe.

luca-della-vedova (commented):
Works great! A few minor comments around

building_sim_plugins/building_ignition_plugins/src/slotcar.cpp
mrushyendra: I think the function just got reordered, which is why Git seems to show it as a new addition. All the battery charging changes have been merged into master already.

building_sim_plugins/building_ignition_plugins/src/slotcar.cpp
```diff
@@ -109,40 +111,76 @@ void SlotcarPlugin::Configure(const Entity& entity,
...
+  if (!ecm.EntityHasComponentType(_entity,
+    components::AngularVelocityCmd().TypeId()))
+    ecm.CreateComponent(_entity, components::AngularVelocityCmd());
+
+  // Keep track of when a payload is dispensed onto/ingested from slotcar
+  // Needed for TPE Plugin to know when to manually move payload via this plugin
+  if (!_ign_node.Subscribe("/item_dispensed", &SlotcarPlugin::item_dispensed_cb,
+    this))
+  {
+    std::cerr << "Error subscribing to topic [/item_dispensed]" << std::endl;
```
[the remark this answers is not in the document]
mrushyendra: It would still be able to subscribe to a topic that hasn't been advertised yet, so the order in which the publishers/subscribers advertise/subscribe to the topic shouldn't matter.

building_sim_plugins/building_ignition_plugins/src/lift.cpp
```diff
@@ -131,9 +176,29 @@ class IGNITION_GAZEBO_VISIBLE LiftPlugin
     auto vel_cmd = ecm.Component<components::JointVelocityCmd>(
       _cabin_joint);
     vel_cmd->Data()[0] = result.velocity;
+
+    // For TPE: Move any payloads that need to be manually moved
+    // (i.e. have a LinearVelocityCmd component that exists)
+    if (std::abs(vel_cmd->Data()[0]) > 0.001)
+    {
+      // TODO: Possibly optimize further by caching positions/entities if required
```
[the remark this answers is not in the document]
mrushyendra: Makes sense! I changed it only search for payloads on a state change.

building_sim_plugins/building_ignition_plugins/src/lift.cpp
```diff
@@ -53,6 +61,35 @@ class IGNITION_GAZEBO_VISIBLE LiftPlugin
...
+      [&](const Entity& entity,
+      const components::Model*,
+      const components::Name* nm,
+      const components::Pose* pose,
+      const components::Static* is_static
+      ) -> bool
+      {
+        // Object should not be static
+        const auto payload_position = pose->Data().Pos();
+        if (!(is_static->Data()) && entity != lift)
```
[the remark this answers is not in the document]
mrushyendra: I tested it, and there isn't currently anything that we want to be fixed and not to move, so I removed the static check.

luca-della-vedova (approved):

luca-della-vedova:
I am very conflicted at whether to control both DART and TPE with `VelocityCmd` or not but this approach is probably the cleanest for now. We can go back to it if there is ever a need to have a more physically accurate simulation in DART.

PR #238, opened 2020-09-28
crowd simulation 2 animation switch

FloodShao:
This PR can be tested under following repo list:
```
repositories:
  rmf/rmf_core:
    type: git
    url: https://github.com/osrf/rmf_core.git
    version: master
  rmf/rmf_schedule_visualizer:
    type: git
    url: https://github.com/osrf/rmf_schedule_visualizer.git
    version: master
  rmf/traffic_editor:
    type: git
    url: https://github.com/FloodShao/traffic_editor.git
    version: fred/switch_anim
  rmf/rmf_demos:
    type: git
    url: https://github.com/osrf/rmf_demos.git
    version: master
  meng_core:
    type: git
    url: https://github.com/osrf/menge_core.git
    version: master
  building_crowdsim_pr_test:
    type: git
    url: https://github.com/FloodShao/building_crowdsim_pr_test.git
    version: master
```
After workspace build process, you can go to `src/building_crowdsim_pr_test/maps/` to find the corresponding launch files. The models used in this test are included in `src/building_crowdsim_pr_test/models`, please copy this two models into `~/.gazebo/models` before launching the simualtion.


building_sim_plugins/building_plugins_common/include/building_sim_common/crowd_simulator_common.hpp
```diff
@@ -132,6 +132,7 @@ class ModelTypeDatabase
     std::string file_name;
     AgentPose3d pose;
     std::string animation;
+    std::string idle_animation = "";
```
luca-della-vedova, line 135: Nit, the default constructor for `std::string` should be enough.

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
```diff
@@ -384,33 +398,65 @@ void CrowdSimulatorPlugin::_update_internal_object(
...
+  }
+  else
+  {
+    anim_name_comp->Data() =
+      _crowd_sim_interface->_model_type_db_ptr->get(obj_ptr->type_name)
+      ->animation;
+    anim_time_comp->Data() +=
+      std::chrono::duration_cast<std::chrono::steady_clock::duration>(
+      std::chrono::duration<double>(distance_traveled / animation_speed));
+  }
```
luca-della-vedova, line 450: Can this be refactored into a state machine? If we keep track of the state we can make it easier to add more states (i.e. transitions), we can also avoid doing operations that are only needed at state change in every loop iteration (such as updating the animation name, locking the yaw for the idle state)

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
```diff
@@ -384,33 +398,65 @@ void CrowdSimulatorPlugin::_update_internal_object(
...
-    exit(EXIT_FAILURE);
-  }
-  auto anim_time = anim_time_comp->Data() +
-    std::chrono::duration_cast<std::chrono::steady_clock::duration>(
-    std::chrono::duration<double>(distance_traveled / animation_speed));
-
-  *anim_time_comp = ignition::gazebo::components::AnimationTime(anim_time);
+  ecm.SetChanged(entity,
+    ignition::gazebo::components::AnimationName::typeId,
+    ignition::gazebo::ComponentState::OneTimeChange);
```
luca-della-vedova, line 459: Moving into a state machine should make it possible to call this instruction only once on state transition instead of every iteration.

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
```diff
@@ -302,6 +302,19 @@ void CrowdSimulatorPlugin::_config_spawned_agents(
     *animation_name_comp = ignition::gazebo::components::AnimationName(
       animation_name);
   }
+  // check idle animation name
+  auto actor_comp =
+    ecm.Component<ignition::gazebo::components::Actor>(entity);
+  for (auto idle_anim : _crowd_sim_interface->get_switch_anim_name())
```
luca-della-vedova, line 308: Nit, I'd probably make `idle_anim` a `std::unordered_set<std::string>` so you can just call the `find` function instead of having to iterate over the whole data structure.

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
```diff
@@ -87,8 +87,8 @@ void CrowdSimulatorPlugin::PreUpdate(
   std::chrono::duration<double> delta_sim_time_tmp = info.simTime -
     _last_sim_time;
   double delta_sim_time = delta_sim_time_tmp.count();
-  if (_crowd_sim_interface->get_sim_time_step() > delta_sim_time)  // not reach one time sim update
-  {
+  if (_crowd_sim_interface->get_sim_time_step() - delta_sim_time > 1e-6)
+  { // not reach one time sim update
```
luca-della-vedova, line 91: Curiosity, why is this needed?

luca-della-vedova (changes requested):
Works for me! A few comments on the logic. Most of the comments on the ignition plugin side also apply to the gazebo one but I avoided repeating them (like the data structure / state machine logic).

luca-della-vedova:
Also can you update the changelogs for the packages with an entry for this PR?

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
FloodShao: Emm, this `std::vector<std::string> _switch_anim_name` contains all the idle animation supported, like "idle", and "stand". In this for loop, we are iterate over {"idle", "stand"} on each Actor object, to find whether the actor sdf defines one of the idle animation. So if we use the unordered_set, we would probably still need to iterate the idle_animation.
BTW, this iteration only happens once for each actor when spawn and initialize the actor. If the idle animation is defined, this idle name will be stored, if not, idle name would be empty.

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
FloodShao: This controls the menge update sim step. 
The logic is 
```
if (delta_sim_time ==  update_time_step)
{
  delta_sim_time = update_time_step
} else
{
 delta_sim_time = 0.0;
}
```
You can't directly use == to check whether 2 doubles are equal, so I used this expression.
Also, this was to make the code consistent with the gazebo version. Because this expression was adopted in gazebo version.

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
luca-della-vedova: Yea it's not really a bottleneck but it's just that if instead of being a vector it was an unordered_set the code would be quite a bit shorter (no need for a for loop, just a find call and a check on its result).

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
luca-della-vedova: I'm having a closer look at the plugin logic and I'm not sure I understand, why are we calling `_update_all_objects` even when we are not in a simulation time step? From what I see in the function the only thing that is done when we are not at the update time step is [update the external agent](https://github.com/osrf/traffic_editor/blob/ee6a40920d637501974da81d3ab12764b8f78e25/building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp#L364-L365), but does updating the external agents pose at a very high frequency (say 1000 Hz) make any difference when the simulator itself only computes steps at 10 Hz?

FloodShao: OK, fix the double comparison. Also checked the `delta_time` and `delta_sim_time` logic, and test shows no difference on updating external agent quite at a high frequency (I think initially I thought updating the external agent faster will make the motion planning more accurate) So I deleted the necessary `delta_time` logic, only use `delta_sim_time` at 0.1s to update all objects.

building_sim_plugins/building_ignition_plugins/src/crowd_simulator.cpp
luca-della-vedova: Oops upon reading the code again I don't think it's possible to do it with a set, apologies for the noise!

luca-della-vedova (approved):

PR #241, opened 2020-10-02
Fixing vertices incorrectly generated after quitting previous edge drawing

MakinoharaShouko:
There is a bug mentioned in [this issue](https://github.com/osrf/traffic_editor/issues/78) that adding edges after cancelling the previous drawing with right click would create an extra vertex. I also found that pressing esc key would have the same effect. This PR has a fix to that issue. So far it is working well when I test it at my end.

### DIFF
--- a/traffic_editor/gui/building.cpp
+++ b/traffic_editor/gui/building.cpp
@@ -149,6 +149,19 @@ bool Building::save_yaml_file()
   return true;
 }
 
+void Building::remove_vertex(int level_index, int vertex_index) {
+  if (level_index >= static_cast<int>(levels.size())
+    || vertex_index >= static_cast<int>(levels[level_index].vertices.size()))
+    return;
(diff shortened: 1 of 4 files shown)


MakinoharaShouko:
I am already checking all the edges to see if the vertex is connected to any of them. If so, the vertex would not be removed. Now I just realized that I should also check the polygons

MakinoharaShouko:
OK, I see. Yeah, it should always be the last vertex because when we do the cancelling it was always the newly added one we need to erase. I will go and fix that.

MakinoharaShouko:
So I decided to used a boolean variable called `added_vertex` to show if the last vertex was added in drawing the edge. There may be a situation in which the user first adds an unconnected vertex and then draw an edge from it then cancel it. Without a boolean value telling us if the vertex was added in drawing the edge, the editor would delete that last vertex by mistake.

PR #243, opened 2020-10-06
Adding actions for plugins + update workflows to foxy

luca-della-vedova:
This PR adds build actions for gazebo and ignition plugins, as well as the menge crowd simulation library. It also updates the workflows to foxy and 20.04.

Sadly menge has _a lot_ of compile warnings which pollutes the build output of the actions but not sure what to do there (we could suppress warnings probably).


luca-della-vedova:
Unsure why only one of the style action failed (!) I suspect it is unrelated since they both ran on the same commit.

PR #245, opened 2020-10-12
only build plugin support if ignition-plugin is present

codebot:
This PR adds a bit of CMake logic and some `#ifdef` throughout the code so that it doesn't compile support for "traffic-editor simulation" plugins if `libignition-plugin` is not present. This will remove the previous hard dependency on `ignition-plugin` and make it a soft dependency, and since nobody is currently using that plugin interface, that seems like a good tradeoff. Hopefully now it will become possible to release this package as a binary, using the normal ROS package release flow.


luca-della-vedova (approved):
Green CI :heavy_check_mark:

PR #249, opened 2020-10-13
Uncommented code block which saves lift parameters when OK is clicked, added initial_floor_name

aaronchongth:
* Bug occurs when default values in combo boxes are used, therefore nothing gets saved into the lift's internal states. The only saving mechanism was left to the `onTextChanged` signal
* Solves https://github.com/osrf/traffic_editor/issues/247


codebot (approved):
Thanks for chasing this down

PR #250, opened 2020-10-14
Enable lifts to work with TPE

mrushyendra:
Modifies lift plugin to issue either joint or model velocity commands based on the physics engine used.


luca-della-vedova (approved):
The code itself looks good, there is no way to test it yet though since link level commands in TPE are still a work in progress.
I guess we would either need a simple test world with lifts that have a single door (maybe a branch on a `rmf_demos` fork?) or we would need to wait for the feature to make it in `ign-physics` first.
What do you think?

mrushyendra:
I was able to test just the lift movement on TPE (minus the door opening/closing) with the lift panel tab on Rviz. Combined with the full testing with DART, would that be sufficient? If not, I can try creating a demo with a lift that has a single door.

luca-della-vedova:
Good point, I managed to test it through full testing in DART and partial testing (manually placing the robot inside the lift) through RVIZ as you mentioned and it looks great!

PR #251, opened 2020-10-14
Fix xml namespace not appearing in generated world files

### LINKED ISSUE #246: Generated world file missing <xmlns:rmf="rmf"> from template
I tried to parse the generated gazebo world file through ElementTree and got this error
```bash
xml.etree.ElementTree.ParseError: unbound prefix
```
My code is
```python3
from xml.etree import ElementTree
tree = ElementTree.parse("<path-to>/<name>.world")
```
It doesn't recognize the `rmf` namespace. Although the namespace line exists in [template file gz_world.sdf](https://github.com/osrf/traffic_editor/blob/3b724bfc7f918e008ebec8ae317da782994301ab/building_map_tools/building_map/templates/gz_world.sdf#L3), but it doesn't appears in my generated world file (an example will be `build/test_maps/maps/door_madness.world`).

This doesn't break Gazebo or Ignition, tinyxm2 seems to ignore this, but it's a small bug that prevents anyone who wants to do some post processing of the generated map files using etree like me.

mrushyendra:
ElementTree's `parse()` method does not store the `xmlns` attribute as part of an element's `attrib` dict, which means
it gets ignored when the template file is parsed and a new world file is generated. This PR inserts the xmlns attribute where
appropriate during the world sdf file generation, since the namespace is only needed for particular elements anyway.

Fixes #246

### DIFF
--- a/building_map_tools/building_map/building.py
+++ b/building_map_tools/building_map/building.py
@@ -200,7 +200,10 @@ class Building:
         tree = parse(template_path)
         sdf = tree.getroot()
 
+        xmlns = {'name': 'rmf', 'value': 'rmf'}  # For any custom elements
+
         world = sdf.find('world')
+        world.attrib['xmlns:' + xmlns['name']] = xmlns['value']
 
         for level_name, level in self.levels.items():
             level.generate_sdf_models(world)  # todo: a better name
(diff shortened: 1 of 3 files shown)


luca-della-vedova:
I did a quick fix a few days ago in #253 that gets rid of the namespace altogether, can you check if it fixes this issues as well? Do you think this would be a better fix?

mrushyendra:
It looks like the issue you raised and #246 is due to the same root cause. Both PRs seem to work well in this case. [This](http://sdformat.org/tutorials?tut=custom_elements_attributes_proposal) page seems to indicate that namespaces may be required, but in practice it seems like the actual libsdformat library has no issues with either.

Briancbn:
Yes, #253 Fixed my problem. Simply removing the namespace solved the problem then.

PR #252, opened 2020-10-14
Fixed RobotMode

Yadunund:
This PR ensures that the RobotMode in the RobotState message published by the slotcar remains as `MODE_DOCKING` until another ModeRequest is received.


aaronchongth (approved):
This looks good! Thanks for putting it together!

PR #253, opened 2020-10-15
Fix namespace for rmf charging plugin

luca-della-vedova:
Using colons for in tags creates issues when parsing XML files in Python, since the colon implements namespaces in the specification, for example while parsing a world I got the following error:

```
xml.etree.ElementTree.ParseError: unbound prefix: line 5819, column 4
```

Changing them to underscore fixes the issue.


Yadunund (approved):
Thanks for the fix!

PR #255, opened 2020-10-16
Feature/remove plugin option

aaronchongth:
* Door edges now have a plugin parameter, which can be selected as `normal` or `none`, this will include or remove the plugin tags when generating the door in the `.world` file
* Lifts have radio buttons to choose whether plugins are required as well, similarly it is read to determine whether plugins are required
* When lifts are selected to not have plugins, the lift models will be set to static.

### DIFF
--- a/building_map_tools/building_map/lift.py
+++ b/building_map_tools/building_map/lift.py
@@ -7,7 +7,7 @@ from .utils import lift_material, visual, collision, box_link, joint
 
 
 class LiftDoor:
-    def __init__(self, yaml_node, name, lift_size, gap):
+    def __init__(self, yaml_node, name, lift_size, gap, plugin=True):
         self.name = name
         self.door_type = yaml_node['door_type']
         # x & y coordinates are with respect to the centre of the cabin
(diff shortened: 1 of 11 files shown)


building_map_tools/building_map/lift.py
```diff
@@ -392,35 +401,39 @@ def generate_cabin(self, world_ele, options):
...
+            plugin_ele.set('name', 'lift')
+            plugin_ele.set('filename', 'liblift.so')
+
+            lift_name_ele = SubElement(plugin_ele, 'lift_name')
+            lift_name_ele.text = f'{self.name}'
+
+            for level_name, door_names in self.level_doors.items():
+                floor_ele = SubElement(plugin_ele, 'floor')
+                floor_ele.set('name', f'{level_name}')
+                floor_ele.set('elevation', f'{self.level_elevation[level_name]}')
```
codebot, line 415: looks like the Python style checker says this line is 2 chars too long. If you break this one at the open-paranthesis, I think it will fit, and then the CI will show green again :heavy_check_mark:

codebot (changes requested):
Thanks @aaronchongth for all of this! Looks like there is just one super minor thing left from the style checker before it's green again. Big picture wise, this looks great to me and I can totally see why this would be beneficial in large worlds where we don't really need controllers running on all the doors we generate, since many doors in large facilities aren't automatic anyway :door:

building_map_tools/building_map/lift.py
aaronchongth: Apologies :face_palm:, I only checked the style locally and saw a whole bunch of divergences from different packages. Thanks for the heads up!

codebot (approved):
awesome thanks!

PR #259, opened 2020-11-02
speedup gz world

tanyouliang95:
Similar to this in ign: https://github.com/osrf/traffic_editor/pull/248, update default max stepsize to speedup gz sim world

### DIFF
--- a/building_map_tools/building_map/templates/gz_world.sdf
+++ b/building_map_tools/building_map/templates/gz_world.sdf
@@ -7,6 +7,15 @@
       <grid>false</grid>
     </scene>
 
+    <physics type="ode">
+      <max_step_size>0.01</max_step_size>
+      <ode>
+        <solver>
+          <iters>30</iters>
+        </solver>
+      </ode>
+    </physics>
+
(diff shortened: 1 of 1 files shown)


Briancbn:
I am actually configuring this myself through a post processing script, will it be better if this is make configurable through building_map_tools cli arguments? or better still in the yaml.

Just a suggestion.

Yadunund:
I too feel this value should be configured by users after world generation as it is subjective to respective hardware configurations.

tanyouliang95:
Since this is just a default value, users can still change it on the gazebo GUI side panel. Having a default just make it slightly convenient to launch and test directly in a speed up manner (reduce the number of clicks). Regardless, any suggestions in making it convenient would be fine.

Briancbn:
> Since this is just a default value, users can still change it on the gazebo GUI side panel. Having a default just make it slightly convenient to launch and test directly in a speed up manner (reduce the number of clicks). Regardless, any suggestions in making it convenient would be fine.

Noted. Fair enough as well.

PR #260, opened 2020-11-03
Added ament exports for crowd simulation common

luca-della-vedova:
Fixes symbol lookups at runtime as reported [here](https://github.com/osrf/rmf_demos/pull/152#issuecomment-720369963)


tanyouliang95 (approved):
Works now. Thanks!

PR #261, opened 2020-11-04
Debug/slotcar battery and mode

Yadunund:
* Updated `mode_request_cb()` to exit if `msg->robot_name` does not match model name.
* Added upper limit to kinematic properties calculated in `compute_discharge()`


tanyouliang95 (approved):
tested, good to go

PR #262, opened 2020-11-05
Migrate ignition plugins to ament_target_dependencies

luca-della-vedova:
Fixes library lookup errors in ignition demos reported in [rmf_demos](https://github.com/osrf/rmf_demos/pull/152#issuecomment-721791154)


Yadunund (approved):
Thanks for addressing the linker error! I'm able to run the `rmf_demos` worlds in ign.

PR #263, opened 2020-11-18
check crowd sim impl pointer before deref

codebot:
During new project creation, the `crowd_sim_impl` pointer will be `nullptr` so the automatic save during creation explodes. This one-liner makes it not explode.


mxgrey (approved):
Working great now, thanks!

PR #264, opened 2020-11-19
idle mode when slotcar is stationary

tanyouliang95:
tested with office world, not breaking.

- pub idle mode
- fix by increase charging radius


Yadunund (approved):
LGTM!

PR #267, opened 2020-12-04
Add pausing feature

mxgrey:
This PR introduces a new pause feature which is needed by https://github.com/osrf/rmf_core/pull/226.

It also fixes a few bugs that have been lingering in the slotcar plugin for some time now.

### DIFF
--- a/building_sim_plugins/building_plugins_common/src/slotcar_common.cpp
+++ b/building_sim_plugins/building_plugins_common/src/slotcar_common.cpp
@@ -87,6 +87,18 @@ double compute_friction_energy(
   return f * m * g * v * dt;
 }
 
+SlotcarCommon::SlotcarCommon()
+{
+  // Make sure we initialize this message to TYPE_RESUME, or else the robot
+  // might just sit and wait around unintentionally.
+  pause_request = rmf_fleet_msgs::build<rmf_fleet_msgs::msg::PauseRequest>()
+      .fleet_name("")
+      .robot_name("")
+      .mode_request_id(0)
+      .type(rmf_fleet_msgs::msg::PauseRequest::TYPE_RESUME)
+      .at_checkpoint(0);
+}
+
 rclcpp::Logger SlotcarCommon::logger() const
 {
   return rclcpp::get_logger("slotcar_" + _model_name);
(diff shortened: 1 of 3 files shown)


building_sim_plugins/building_plugins_common/src/slotcar_common.cpp
```diff
@@ -87,6 +87,18 @@ double compute_friction_energy(
   return f * m * g * v * dt;
 }
 
+SlotcarCommon::SlotcarCommon()
+{
+  // Make sure we initialize this message to TYPE_RESUME, or else the robot
+  // might just sit and wait around unintentionally.
+  pause_request = rmf_fleet_msgs::build<rmf_fleet_msgs::msg::PauseRequest>()
+      .fleet_name("")
```
arjo129, line 95: Use of 4 spaces as opposed to 2 spaces.

building_sim_plugins/building_plugins_common/src/slotcar_common.cpp
```diff
@@ -186,11 +207,10 @@ void SlotcarCommon::path_request_cb(
 
     Eigen::Quaterniond quat(
       Eigen::AngleAxisd(msg->path[i].yaw, Eigen::Vector3d::UnitZ()));
-    trajectory[i].translation() = v3;
+    trajectory.at(i).translation() = v3;
     trajectory[i].linear() = Eigen::Matrix3d(quat);
```
arjo129, line 211: Since we are using `at` for indexing, for consistency it may be better if we use `at` throughout? I'm presuming we are using `at` here due to its safe access?

mxgrey: > I'm presuming we are using at here due to its safe access?

Correct.

In order to reach line 211, the `trajectory.at(i)` of the previous line must have succeeded without throwing an exception, so line 211 is guaranteed to have an element with index `i`, making the bound check redundant.

That being said, the bound check of `.at(~)` is negligible overhead compared to everything else that's going on in the simulation, so I don't mind using it redundantly if others would prefer that.

Truthfully, neither of them should require an `.at(~)` since they were both explicitly resized earlier in this function, but I was experiencing some undefined behavior related to the trajectories, and so I put in this `.at(~)` to help with debugging it. I think the undefined behavior was related to multiple read/write issues, which should now be resolved by the new mutex, but I don't mind keeping the `.at(~)` here and adding one to line 211 just in case.

mxgrey:
Style issues are fixed now :+1:

arjo129 (approved):
Looks great! Had no issues while testing.

PR #268, opened 2020-12-04
Single level crowdsim remake

tianenchong:
Works with new building.yaml format (read and write).
Crowdsim UI remake.


luca-della-vedova:
Can you give us more details about what changed, in what way and what this PR addresses?

tianenchong:
Originally, in building yaml file, crowdsim tag was a separate entity beside building. This pr brings crowdsim tag to under L1 (level 1) so that crowdsim settings are independent across each level. This serves as a preparation for independent multilevel crowdsim. Only L1 is supported at the moment. All non-static models for human are added under models tag, where originally they fall under crowdsim tag and dynamically generated and inserted to the model during runtime. Various internal crowdsim child tag renaming and some crowdsim properties trimming are done to reduce redundancy (when a particular value can be just referred from another child tag).

**A word of caution: this pr has only been tested with single level map. There might be some unforeseen behaviour in multilevel map.** Multilevel map will be tested in the next iteration of multilevel independent crowdsim.

Various python generation scripts are updated to accommodate new changes in building yaml file (see "Files changed").

Traffic editor UI sees new human properties tag specific to each vertex, so a user can select a vertex and add simulated human under the properties. In the past, a user have to manually specify coordinate for a particular vertex in the crowdsim configuration table.

Model type table and agent profile table are combined so that agent names - common to both tables - are edited in a single place.

Agents table are replaced with just a single text field for the external agent names (with default setting for external agent remain hidden to user), because the internal agents (simulated human) can be directly manipulated by selecting a particular vertex and changing its human properties.

tianenchong:
Superseded by multilevel crowdsim pr.

PR #271, opened 2020-12-11
Remove Slotcar and Lift AAB components when not required to improve Ign Demos rtf

mrushyendra:
Reads or stores the `AxisAlignedBox` component in the Lift and Slotcar plugins before removing the component. Improves rtf when running Ignition demos since bounding box computation is expensive. 

We need to read this value beforehand instead of only creating the component when we need it because it usually takes at least one more update/preupdate iteration to compute the `AxisAlignedBox`.


luca-della-vedova (approved):
Tested and performance is much better, as well as delivery tasks still work as expected :+1:

PR #272, opened 2020-12-11
fix to compilation when no opencv

marcoag:
Traffic_editor wont compile when opencv is not installed but the ignition plugin is there. This will correct the compilation error.


codebot (approved):
Thanks.

PR #274, opened 2020-12-15
Fix undefined initialization of _old_ang_vel and _old_lin_vel

mrushyendra:
`old_lin_vel` and `old_ang_vel` were not set upon initialization, causing the first iteration of the discharge calculation to output undefined values.


Yadunund (approved):
Thanks for addressing this bug!

Issue #275, opened 2020-12-17
Snapping to waypoints should take the user's zoom into account


mxgrey:
A nifty feature of the traffic editor is that it gives some wiggle room when clicking to place waypoints, where it will snap to existing waypoint if it's close enough to where you're clicking. This is helpful to ensure that graphs are connected the way they're supposed to be.

However, it seems that this snapping feature does not take the user's level of zooming into account. As a result, if I'm zoomed in very closely (which is necessary when tracing very large maps), the traffic editor will snap my click to a waypoint that's actually pretty far from where I'm clicking. This forces me to click somewhere very far away to create a new waypoint and then move the waypoint into place.

codebot:
closed by #276

PR #276, opened 2020-12-17
scale add-lane vertex snap distance correctly

codebot:
Addresses #275  at least for now, needs to be combined with the Undo work though after that PR is in.

### DIFF
--- a/traffic_editor/gui/editor.cpp
+++ b/traffic_editor/gui/editor.cpp
@@ -1837,9 +1837,18 @@ void Editor::mouse_add_edge(
       return;
     }
 
+    const double vertex_radius_meters = 0.1;
+    const double vertex_dist_thresh =
+      project.building.levels[level_idx].drawing_meters_per_pixel /
+      vertex_radius_meters;
+
     const int prev_clicked_idx = clicked_idx;
     clicked_idx = project.building.nearest_item_index_if_within_distance(
-      level_idx, p_aligned.x(), p_aligned.y(), 10.0, Building::VERTEX);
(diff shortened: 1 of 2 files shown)


mxgrey (approved):
Seems to work as intended!

traffic_editor/gui/editor.cpp
```diff
@@ -1837,9 +1837,18 @@ void Editor::mouse_add_edge(
       return;
     }
 
+    const double vertex_radius_meters = 0.1;
+    const double vertex_dist_thresh =
+      project.building.levels[level_idx].drawing_meters_per_pixel /
+      vertex_radius_meters;
```
mxgrey, line 1843: After trying this on a different map, I think this fraction is inverted. It seems like it should really be

```
vertex_radius_meters / project.building.levels[level_idx].drawing_meters_per_pixel;
```

mxgrey (changes requested):

traffic_editor/gui/editor.cpp
codebot: I'm so embarrassed. I have no idea why I thought this was working (I guess just coincidence?), but it "felt" like it did on the test map... :face_with_head_bandage: but yes, we want the threshold to be calculated in pixels, and inverting the fraction as you propose does that.

codebot: I'll test it on a few maps of wildly different scales right now and update the PR.

mxgrey (approved):

PR #281, opened 2021-01-12
Switch to PeriodicChange for ignition crowd simulation plugins

luca-della-vedova:
The use of `OneTimeChange` for components in ignition causes the scene broadcaster to serialize and publish the whole state, incurring in a major performance loss when using the ignition crowd simulation plugin, as documented in [ign-gazebo](https://github.com/ignitionrobotics/ign-gazebo/pull/544).

This PR changes the components state to `PeriodicChange`, to be merged after the matching [ign-gazebo PR](https://github.com/ignitionrobotics/ign-gazebo/pull/544) (or to be updated if the issue is fixed in a different way).


luca-della-vedova:
Closing at the moment to prepare for repo migration

PR #289, opened 2021-02-03
remove unused variables, fix delete crash, initialize to nullptr

ddengster:
to repro: press T and then M, you should get a crash due to deleting un-initialized pointers. also this PR cleans up some unused variables


codebot (approved):
Crash reproduced and fix verified. Thank you for chasing this down and for the clean PR. :+1:

PR #290, opened 2021-02-03
place orientation icon at correct Z stack so it's visible again

codebot:
When we started to render using proper Z stacks, unfortunately the orientation icons were not assigned a Z height, so they were below everything. Fixed now.


Yadunund (approved):
Tested and the orientation icons are properly rendered above the lanes!

PR #291, opened 2021-02-03
building_map_tools feature: building_map_combiner

codebot:
This PR introduces a new Python program, `building_map_combiner` and implements its functionality in the `building_map` Python package. It allows you to add vertices and traffic lanes from one map into another map. As the vertices are being added, they will be transformed into the same coordinate space using fiducials, provided there exist at least 3 equally-named fiducials in each map. This was implemented primarily by refactoring the existing fiducial-alignment code used for multi-floor world generation into a form that could be used for multi-map alignment.

In order to implement this feature, it was also necessary to add map-writing `to_yaml()` functions throughout the `building_map` Python package. It would be nicer if the YAML flow/block style toggling was happening the same way in this Python writing implementation as it currently does in C++, but I'll add that in a separate PR later.

### DIFF
--- a/building_map_tools/building_map_combiner/building_map_combiner.py
+++ b/building_map_tools/building_map_combiner/building_map_combiner.py
@@ -8,16 +8,16 @@ from building_map.building import Building
 
 def main():
     parser = argparse.ArgumentParser(
-        prog="building_map_combiner",
-        description="Combine two .building.yaml files"
+        prog='building_map_combiner',
+        description='Combine two .building.yaml files'
     )
 
-    parser.add_argument("PRIMARY_INPUT", type=str,
-                        help="Primary building.yaml file")
-    parser.add_argument("SECONDARY_INPUT", type=str,
-                        help="Secondary building.yaml file")
-    parser.add_argument("OUTPUT", type=str,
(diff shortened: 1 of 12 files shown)


building_map_tools/building_map_combiner/building_map_combiner.py
```diff
@@ -0,0 +1,43 @@
...
+import argparse
+import os
+import yaml
+
+from building_map.building import Building
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        prog="building_map_combiner",
```
aaronchongth, line 11: nitpick: using double quotes while code base uses single quote

aaronchongth (approved):
The refactor looks good! Are there certain files that you have been testing this with?

Just a side note, I noticed that we are inverting the y coordinates back on the `to_yaml` function for `Model`, `Vertex`, and `Lift`, but not for `LiftDoor` and `Fiducial`, is that intended? I haven't run any tests, but just as a sanity check I thought I'd ask

building_map_tools/building_map_combiner/building_map_combiner.py
codebot: Thank you :bow:  fixed now.

PR #292, opened 2021-02-04
change vertex shape for parking/holding points

codebot:
This PR changes the shape of vertex icons to indicate if they are set to be holding and/or parking points in the navigation graph.

 * holding points will be rendered as slightly larger squares
 * parking points will have an outline of an even-larger square
 * otherwise the vertex will be rendered as a circle, as before.

### DIFF
--- a/traffic_editor/gui/building_level.cpp
+++ b/traffic_editor/gui/building_level.cpp
@@ -979,8 +979,7 @@ void BuildingLevel::draw(
   for (const auto& v : vertices)
     v.draw(
       scene,
-      vertex_radius / drawing_meters_per_pixel,
-      QColor::fromRgbF(0.0, 0.5, 0.0));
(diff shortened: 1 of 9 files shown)


Yadunund (approved):
Tested it out and works great! The maps in traffic editor now convey more information visually. It would be nice also if waypoints with `is_charger` property set `true` could visually indicate that they are charging waypoints. Maybe a different color infill? Just a thought for a future PR :)

Yadunund:
The ring concept is fantastic! The waypoint attributes are a lot more intuitive now without any legend! I noticed that the stopwatch icon gets rendered at `-45deg` which tends to overlap with the waypoint name (see below)
image

Perhaps setting the bearing to `225deg` is better?
image

Yadunund:
Awesome! It's good to merge 👍🏼

PR #294, opened 2021-02-12
Add UI for adding correspondence points

gbiggs:
This adds the required UI to be able to specify correspondence points for each layer of a level, then save those correspondence points to a separate YAML file that can be used with the map transformer library.


gbiggs:
I've renamed the correspondence points tool to "layer alignment points". I haven't renamed the fiducial tool. Do you feel strongly that it should be renamed as well?

codebot (approved):
Works for me on my test case, and passes all CI checks.

PR #298, opened 2021-02-22
correctly exit add-edge mode, fix #297

### LINKED ISSUE #297: Unwanted lanes added when adding a new lane
Running into this issue when creating lanes. Best explained by the video attached.
In the video below, I first add a lane between the waypoint on the left and the one on the right. I then hit `Esc` and then try adding a lane from the bottom waypoint to the left one. But when I click the bottom waypoint, a lane from the right waypoint to the bottom one gets added. It's almost as if the `add_lane` tool does not reset when `Esc` is hit. Only the rendering of the blue preview resets.

https://user-images.githubusercontent.com/13482049/108651285-66c99680-74fc-11eb-949d-deeb8088f58a.mp4

codebot:
The add-edge tools were correctly exiting via right-click, but not via `[ESCAPE]` key. This PR just switches the order the underlying functions were being called to fix it. It also fixes an unrelated crash that resulted from exiting the tool correctly.

### DIFF
--- a/traffic_editor/gui/editor.cpp
+++ b/traffic_editor/gui/editor.cpp
@@ -1028,9 +1028,9 @@ void Editor::keyPressEvent(QKeyEvent* e)
       break;
     case Qt::Key_S:
     case Qt::Key_Escape:
-      tool_button_group->button(TOOL_SELECT)->click();
       project.clear_selection(level_idx);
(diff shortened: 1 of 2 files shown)


Yadunund (approved):
Works perfectly! Thanks for the fix 🙇🏼

PR #299, opened 2021-02-25
allow parsing failures to not crash building_crowdsim

codebot:
Previously the `building_crowdsim` generator would throw Python exceptions (crash) if it didn't find the YAML keys it expected in the building files. This was not convenient if using a `cmake GLOB` to collect all the building files and generate their maps, since it would lead to broken builds if any of the buildings didn't have any human data. This PR tweaks `building_crowdsim` a bit so that it doesn't crash in this case and just exits cleanly without doing anything.


Yadunund (approved):
Thank you for this fix 🙇🏼

PR #300, opened 2021-02-26
Add minimal CI test

codebot:
This PR reworks the `cmake` of the `traffic_editor` package to make it testable, and creates a minimal QTest:
 * move the GitHub workflows to `foxy` (they were on `eloquent` before)
 * create a static library for all of the sources other than `main.cpp`
 * builds the existing GUI by linking `main.cpp` with that static library (no change for users)
 * creates a minimal QTest, `test_gui.cpp` also with that static library, and register it in `ament_cmake` so that it runs with `colcon test` and dumps its output in (I think) a workable format and place. At least it seems to work as expected when I create a failing `QCOMPARE()` in QTest, reporting the failure as expected with all the nice QTest log info.
 * the minimal QTest just instantiates the `Editor` class and then deletes it, using the offscreen renderer so that it can run on GitHub CI. This doesn't really test anything, other than just making sure that the `Editor` constructor doesn't explode.
 * renames the `build` workflow to `ci` and actually runs `colcon test` and `colcon test-result` on it
 
The idea is that we can use this as a starting point to exercise the GUI in QTest, emulate GUI actions and clicks, and start getting some code coverage.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,9 +2,9 @@ name: ci
 on: [push, pull_request]
 jobs:
   ci:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container:
-      image: docker://ros:eloquent-ros-base-bionic
+      image: docker://ros:foxy-ros-base-focal
     
     steps:
     - name: osrf-repo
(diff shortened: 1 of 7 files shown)


Yadunund (approved):
This is awesome. Excited to see `testGui()` updated with more tests in the near future!

PR #301, opened 2021-03-03
Remove project abstraction layer and built-in custom simulation plugin interface

codebot:
Fairly massive cleanup of lots of things in the GUI. Lots of deferred maintenance as well as scope-tightening to return the scope of `traffic-editor` to just traffic editing:
 * remove the "simulation plugin" interface. It was half-baked and is better implemented elsewhere in the ecosystem
 * as a consequence, remove the `Project` abstraction and return the `Building` to its previous position as the top-level document. This involved a lot of refactoring.
 * remove `.project.yaml` files, since they are no longer needed. The document file is just the `.building.yaml` files again.
 * remove the duplicated CI that was previously needed for the "with-ignition-plugin" and "without-ignition-plugin" cases
 * external traffic map files are no longer in scope, so remove their placeholder implementation (it never worked) as well as the Editor Modes that were intended to be needed, that was why you had to toggle between editing the building and traffic maps, even though the "external traffic map file" system was never fully implemented.
 * now that the plugin interface is gone, move all headers back to being internal (non-installed) headers, to emphasize that they are not intended as an external API for traffic-editor functionality.
 * fix a few crashes related to bounds checking of correspondence points


traffic_editor/gui/actions/move_correspondence_point.h
```diff
@@ -15,19 +15,19 @@
  *
 */
 
-#ifndef ACTIONS__MOVE_CORRESPONDENCE_POING_HPP_
-#define ACTIONS__MOVE_CORRESPONDENCE_POING_HPP_
+#ifndef ACTIONS__MOVE_CORRESPONDENCE_POING_H_
+#define ACTIONS__MOVE_CORRESPONDENCE_POING_H_
```
luca-della-vedova, line 19: Nit, typo here

luca-della-vedova (approved):
Looks good! Tried it as well and didn't see any major regression, just a small comment on a typo

traffic_editor/gui/actions/move_correspondence_point.h
codebot: I am very impressed with your :eyes: for detail! Thanks for finding this. Fixing now.

PR #304, opened 2021-03-16
fixes #303, bug in loading/saving relative paths

### LINKED ISSUE #303: building.yaml file not updated when clicking Save
Running latest master on Ubuntu 20.04.

Steps to re-create the problem:
1. Launch `traffic-editor` from the terminal and pass in the path to the `.building.yaml` file
2. Select any waypoint and add new properties to it.
3. `Ctrl+S` to save and close `traffic-editor`
4. Perform Step 1 and select previously modified waypoint. The changes you made previously will be missing

codebot:
Relative paths were not being handled properly when loading and saving. Also, the `ofstream` object was not being checked for validity, so this was a silent error and edits would not be written to disk, resulting in much confusion. Now an error box will pop up if the `ofstream` is not pointing to an actual file, and the relative vs. absolute filename handling is fixed.

### DIFF
--- a/traffic_editor/gui/building.cpp
+++ b/traffic_editor/gui/building.cpp
@@ -162,6 +162,11 @@ bool Building::save()
   YAML::Emitter emitter;
   yaml_utils::write_node(y, emitter);
   std::ofstream fout(filename);
+  if (!fout)
+  {
+    printf("unable to open %s\n", filename.c_str());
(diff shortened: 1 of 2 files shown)


Yadunund (approved):
Thanks for all the effort with diagnosing and fixing this sneaky issue! 🙇🏼

PR #306, opened 2021-03-18
Disable crowd sim without failing

tanyouliang95:
Previously, if `MENGE_RESOURCE_PATH` is not provided, the `crowd_sim` plugin will fail, resulted in failing to launch a demo world. Instead of failing, the minor change here is to disable `crowd_sim` if `MENGE_RESOURCE_PATH` is not provided. 

With this, a subsequent pr will consist of an optional `use_crowdsim` arg (e.g.  `XX.launch.xml use_crowdsim:=1` ) in the demo world launch files.


luca-della-vedova:
Can we also shortcut and early the Update functions if crowdsim is disabled? I don't see any check for it so I'm not sure what they would do

luca-della-vedova (approved):
Nice, thanks for addressing the feedback!

PR #307, opened 2021-03-26
print console warning when trying to load project.yaml file

codebot:
Since the `.project.yaml` file is no longer being used, this PR just adds a (hopefully) helpful console warning if a `.project.yaml` filename is passed on the command line.

### DIFF
--- a/traffic_editor/gui/building.cpp
+++ b/traffic_editor/gui/building.cpp
@@ -52,7 +52,7 @@ Building::~Building()
 /// in the YAML file.
 bool Building::load(const string& _filename)
 {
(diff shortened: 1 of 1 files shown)


traffic_editor/gui/building.cpp
```diff
@@ -55,14 +55,23 @@ bool Building::load(const string& _filename)
   printf("Building::load_yaml_file(%s)\n", _filename.c_str());
```
mxgrey, line 55: Unrelated nitpick:

```suggestion
  printf("Building::load(%s)\n", _filename.c_str());
```

mxgrey (approved):

PR #312, opened 2021-03-29
avoid crash in empty crowdsim save routine

codebot:
In certain conditions (especially when creating new buildings), a crash can be triggered by this line in the crowdsim `to_yaml()` function. Adding this one-line check avoids that situation.


Yadunund (approved):
Looks good!

Issue #311, opened 2021-03-29
Adding a level does not respond


destkk:
Ubuntu 20.04

Hi there. I have tried by adding a new level in traffic editor and importing a png file and giving it a level name. After adding it the image does not appear. Previously it was working fine I could open the project.yaml file from rmf_demos and it is working well. Out of a sudden it does not work for any of the project files.
```
changing directory to [/home/rosuer/rmf_demos_ws/install/rmf_demo_maps/share/rmf_demo_maps/hotel]
Building::load_yaml_file(hotel.building.yaml)
changing directory to [/home/rosuer/rmf_demos_ws/install/rmf_demo_maps/share/rmf_demo_maps/hotel]
parsing level [L1]
parsing level [L2]
used 2 measurements to estimate meters/pixel as 0.05000
used 2 measurements to estimate meters/pixel as 0.05000
transform 0->1: scale = 1.00000 translation = (-54.00, 0.00)
transform 1->0: scale = 1.00000 translation = (54.00, 0.00)
project does not have a sim plugin
```

Does anyone know how to resolve this issue? Is it because the project does not have sim plugin?

The screenshot below is for hotel.project.yaml in rmf_demos
![Screenshot from 2021-03-29 13-33-56](https://user-images.githubusercontent.com/63645940/112791161-7a37c680-9093-11eb-833a-cd377037552c.png)

codebot:
Sim plugin is unrelated and not an issue here. Just checking, so we can home in on the issue: "what if" you add a level, exit, and then re-open traffic-editor -- does that make the level background image appear?

codebot:
Ohhh -- it looks like you haven't built traffic-editor from source in a while (few weeks?) since your screenshot shows a `Project` menu. It should show a `Building` menu now in the upper-left. Can you update your local checkout and rebuild from source?

destkk:
> Ohhh -- it looks like you haven't built traffic-editor from source in a while (few weeks?) since your screenshot shows a `Project` menu. It should show a `Building` menu now in the upper-left. Can you update your local checkout and rebuild from source?

Hi there. May I know how do I rebuild from source and what I should do with the old files? Sorry I'm a newbie in this.

codebot:
How did you currently install the `traffic_editor` version that you are running in the screenshot above?

destkk:
> How did you currently install the `traffic_editor` version that you are running in the screenshot above?

[...]

PR #318, opened 2021-03-29
provide zoom-reset and clamp on scale factor

### LINKED ISSUE #315: It is possible to zoom infinitely out
#311 describes a case where it's possible to zoom infinitely out. Once the zoom factor is set to `0.000000`, no amount of mouse-wheel scrolling can restore the zoom level.

codebot:
Fix #315 and provide a better solution for #311 by providing a zoom-reset menu action (for recovery) and a zoom-clamp calculation in the mouse-wheel handler to prevent the zoom factor from flying to zero, which make it impossible to recover.


Yadunund (approved):
Tested and works great! 💯

Issue #319, opened 2021-03-30
Unable to generate world file and models from 3 levels map


destkk:
I'm trying to generate a .world file and models from the 3 levels map that I have built in traffic editor using this command 
`ros2 run rmf_building_map_tools building_map_generator gazebo  ~/rmf_ws/install/rmf_demos_maps/share/rmf_demos_maps/test_building.building.yaml test_building.world ~/rmf_ws/install/rmf_demos_maps/share/rmf_demos_maps/maps/test_building/models` but I am receiving this error and unable to build the map. Could it be possible that I have built the map wrongly?

```
building name: test_building
parsing level L1
parsing level L2
parsing level L3
calculating levels relative to L1
level L1 scale: 0.049999520301876176
calculating level L2 offset and scale...
  4 common fiducials:
Transform.set_from_fiducials()
  4 fiducial pairs:
    (7814.1, -5614.0) -> (7830.4, -5606.0)   ref1
    (7842.3, -5621.7) -> (7860.4, -5614.8)   ref2
    (7835.4, -5535.4) -> (7852.4, -5521.5)   ref3
    (7861.7, -5543.8) -> (7881.6, -5530.4)   ref4
    calc dist ref1 <=> ref2
    calc dist ref1 <=> ref3
    calc dist ref1 <=> ref4
    calc dist ref2 <=> ref3
    calc dist ref2 <=> ref4
    calc dist ref3 <=> ref4
Bearings:
[(-0.2673258533057868, -0.2847918192598588), (1.3065081154511626, 1.316608516250293), (0.9749057912340271, 0.9755285900765562), (1.6511376819099384, 1.6569610474329988), (1.3266570329172527, 1.324780805386846), (-0.30583893247514277, -0.29509117110947225)]
  -0.017465965954072016
  0.010100400799130327
  0.0006227988425291375
  0.0058233655230603976
  -0.0018762275304067266
  0.010747761365670516
  Circular mean: -0.0013254962167178644
Distances:
[(29.24060498690073, 31.27161556427871), (81.37131959111895, 87.30634064602675), (84.84514390935956, 91.31660153553688), (86.54917570953565, 93.63437260429552), (80.3280698448556, 87.01600818240259), (27.645917329689453, 30.5714307973966)]
mean relative scale: 1.0816019487363226
translation: (29.086, -21.052)
calculating level L3 offset and scale...
  4 common fiducials:
Transform.set_from_fiducials()
  4 fiducial pairs:
[...]

PR #321, opened 2021-03-30
improve usage of Shapely on very complex floor polygons

codebot:
On very complex and large floor polygons (e.g., an entire floor of a large building, with many long and narrow corridors), the recursive triangulation algorithm was exploding. For reasons not fully understood, sometimes the Shapely library returns duplicate points even after calling `simplify()` with the default algorithm. This PR changes the call to Shapely's `simplify()` function to use an alternative algorithm that does not seem to return duplicate points. This PR also increases the simplification tolerance to 0.5 pixel, which still seems more accurate than anybody is doing for hand-annotation. With these changes, it seems to triangulate the map on which it was crashing in #319

### DIFF
--- a/rmf_building_map_tools/building_map/floor.py
+++ b/rmf_building_map_tools/building_map/floor.py
@@ -82,6 +82,7 @@ class Floor:
             if triangulation_debugging:
                 tri_x, tri_y = triangle.exterior.coords.xy
                 plt.plot(tri_x, tri_y, 'k', linewidth=1)
+                plt.pause(0.1)
 
             # create a new geometry by intersecting this triangle with the
             # original polygon. This may create a triangle, or it may also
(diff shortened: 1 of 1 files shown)


Yadunund:
It seems like the changes to the algorithm is resulting in gaps in the flooring as seen in the `office` and `airport_terminal` worlds

image
image
image

rmf_building_map_tools/building_map/floor.py
```diff
@@ -98,17 +99,19 @@ def triangulate_polygon(self, polygon, triangles):
 
             # now we need to actually deal with an intersection area
             if geom.geom_type == 'Polygon':
-                # first ensure the winding is in canonical order (CCW)
-                poly = shapely.geometry.polygon.orient(geom)
+                # simplify it to remove duplicate points, which seems to
+                # happen sometimes
+                poly = geom.simplify(0.5, False)
```
Yadunund, line 104:
```suggestion
                poly = geom.simplify(0.001, False)
```

Yadunund (changes requested):
Managed to get the floors to generate without gaps by decreasing the `simplify()` tolerance back to `0.001`


image

image
image

Yadunund (approved):

PR #322, opened 2021-03-30
avoid crashing when generating undefined floor polygons. cleanup.

codebot:
If a floor polygon had zero, one, or two vertices, building map generation would crash. This PR adds some checks for this case. Also removes some comment blocks and old debugging code.


Yadunund (approved):
Tested and works!

PR #327, opened 2021-03-31
update layer image immediately after OK button

### LINKED ISSUE #324: layer images do not automatically update when image filename is changed
## Steps to reproduce the bug

1. create new layer with an image file
1. click the `Edit...` button next to that layer in the layers table
1. select a new image filename for that layer and click OK
1. observe that the layer image does not change

I suspect that after a save/exit/open cycle, it would be changed, but that's not a great UX :laughing:

codebot:
Immediately re-load layer image after filename is changed and OK is pressed in the layer edit dialog box. Fix #324.

### DIFF
--- a/rmf_traffic_editor/gui/layer_dialog.cpp
+++ b/rmf_traffic_editor/gui/layer_dialog.cpp
@@ -175,6 +175,10 @@ void LayerDialog::ok_button_clicked()
       return;
     }
   }
+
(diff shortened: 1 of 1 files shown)


Yadunund (approved):
The layer did correctly reload if saved/re-opened but it's great to have this improvement to the UX 😄

PR #328, opened 2021-03-31
handle editing multiple layers with same name.

### LINKED ISSUE #325: layers with the same name cannot be edited independently
### Steps to reproduce the bug

1. create a layer named `test` and assign it some properties (offset, filename, etc.)
1. create another layer named `test` and assign it some different properties
1. click the `Edit...` button next to the second layer named `test`
1. observe that the popup dialog shows the properties for the first dialog named `test`

codebot:
Previously, if two layers were added with the same name, it would be impossible to edit the second one, because the layer editing dialog was only searching layers by name, not by index. Now the layer dialog finds the requested layer by index in the layers table. Fix #325 


Yadunund (approved):
Tested and solves the problem!

Issue #22, opened 2019-12-07
click+drag behavior in KDE


codebot:
Report of potential weirdness with click+drag event handling in KDE: the whole window moves when clicking and dragging, as if you press alt+click+drag. Need to test further.

codebot:
click-drag is no longer required to create edges, so presumably this problem is "fixed" for now.

PR #330, opened 2021-04-05
Fix material values for sdf compliance

luca-della-vedova:
### Fixed bug

Similar to [rmf_demos #22](https://github.com/open-rmf/rmf_demos/pull/22), since `libsdformat11` ambient / diffuse values outside of the range dictated by the specification [0.0 - 1.0] cause runtime errors. This makes it impossible to run the scenarios in Ignition Edifice (and later version).

### Fix applied

Scaled the door ambient and diffuse values down to the specification's range

### DIFF
--- a/rmf_building_map_tools/building_map/utils.py
+++ b/rmf_building_map_tools/building_map/utils.py
@@ -20,9 +20,9 @@ def door_material(options):
     material_ele = Element('material')
     # blue-green glass as a default, so it's easy to see
     ambient_ele = SubElement(material_ele, 'ambient')
-    ambient_ele.text = '{} {} {} {}'.format(120, 60, 0, 0.6)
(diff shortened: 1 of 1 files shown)


tanyouliang95:
We might also need to update the ign version in [this](https://github.com/open-rmf/rmf_traffic_editor/blob/24567f68c9071f7d1de83054a0e926239d0c6d5a/rmf_traffic_editor/package.xml#L13) `package.xml`

luca-della-vedova:
> We might also need to update the ign version in [this](https://github.com/open-rmf/rmf_traffic_editor/blob/24567f68c9071f7d1de83054a0e926239d0c6d5a/rmf_traffic_editor/package.xml#L13) `package.xml`

Nice find! That dependency is actually not necessary anymore since the whole ignition plugin based system in `traffic_editor` has been removed with the latest cleanup. I removed all mentions of ignition there

tanyouliang95 (approved):

PR #334, opened 2021-04-09
building_map_server: don't crash when missing image file

### LINKED ISSUE #333: building map server crashes when an image file is not present
Currently `building_map_server` crashes when a referenced image file is not present. It would be nicer to just print a warning and continue, since lots of RMF still works fine (traffic management, etc) even when there is a missing building map file. In the future, perhaps we could generate a placeholder "warning" image so it's clear why the UI's have no background image(s).

codebot:
Currently, `building_map_server` crashes on startup when a building-floorplan image file is missing. Because the floorplan images are only needed for UI's and not "the rest of RMF," this PR changes this crash to an error message in the log. Fixes #333


Yadunund (approved):
Tested and works excellently!

PR #338, opened 2021-04-22
Convert wall textures from 1D to 2D

luca-della-vedova:
### Implemented feature

Walls' textures are now 2D instead of 1D, this allows mapping complex textures to wall and removing the glitches that appeared when some textures (such as tiles) were applied to walls.

To be fully honest I don't understand 100% the obj specification and whether the logic I added for the texture vertices is correct. I tried to follow the same ordering as the polygon vertices and "unwrap" the wall textures manually.

This implementation is only a starting point, followup work should parametrize texture scale for walls, as it happens for floors, as well as adding more wall specific textures. Now we are assuming 2.5m height x 1m width textures.
I split the results in a classic western way.

**The good**, what would happen if we parametrized the texture height and applied the correct scale to a tiles texture (example taken from airport world):

image

**The bad**, the state with the current PR, vertical scale is wrong since the texture is 1x1 but we assume 2.5x1:

image

**The ugly**, what it was before this PR (!)

image

### DIFF
--- a/rmf_building_map_tools/building_map/wall.py
+++ b/rmf_building_map_tools/building_map/wall.py
@@ -61,7 +61,7 @@ class Wall:
 
             # calculate faces for all the wall segments
             wall_verts = np.array([])
-            texture_lengths = [0]
+            texture_lengths = []
 
             # normal #1 is the vertical normal for wall "caps"
             norms = np.array([[0, 0, 1]])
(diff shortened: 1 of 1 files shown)


codebot (approved):
nice! In the future we can add parameters to diverge from the 2.5m x 1m default scaling, but that's a future expansion.

Does the previous default wall texture still look reasonable with this new code that is doing 2d textures (not just 1d) ?

luca-della-vedova:
> Does the previous default wall texture still look reasonable with this new code that is doing 2d textures (not just 1d) ?

Yap! Office demo walls are unchanged

image

PR #339, opened 2021-04-22
Fix/ci package name

luca-della-vedova:
### Fixed bug

CI was not doing anything since `traffic_editor` has been renamed to `rmf_traffic_editor`. It was also installing Ignition packages that are not needed anymore.

### Fix applied

Fixed the github workflow, as well as the tests CMakeLists, so it correctly compiles and tests `rmf_traffic_editor`, also removed the unnecessary Ignition dependencies.


codebot (approved):
AHHH thank you for fixing this!

PR #340, opened 2021-04-23
Automatic alignment of robot-map layers to floorplans

codebot:
### Implemented feature

 * annotate layer-specific features by clicking on the layer and then clicking in the working area
 * define constraints between features
 * automatically compute transformation (rotate/scale/translation) between base floorplan image and layer image (typically robot-generated maps) using `ceres` as the NLLS solver
 * undo/redo defined for the add/delete feature and add/delete constraints

### Cleanup / refactoring
 * combine the `Level` and `BuildingLevel` classes to reduce code lasagna, since (after some cleanup a few months ago) `BuildingLevel` was the only child class of `Level`, so the hierarchy was just making stuff harder to find without adding any benefit.


Yadunund:
This is an amazing feature that many will find super helpful!
In my testing I did encounter one issue. Opening an old `.building.yaml` that already had a robot map layer added and aligned through manual adjustments of the X, Y and Rotations, results in 💥 . More specifically, the Layer Properties get reset. In my case the `Meters per pixel` value defaulted to 1 which resulted in a scaled up map.

image

Yadunund (approved):
Thanks for addressing the concern! Tested and works great with legacy maps 👍🏼

PR #341, opened 2021-04-23
Fix asset path after package renaming

luca-della-vedova:
### Fixed bug

By default, `rmf_traffic_editor` was looking for assets `traffic_editor_assets` folder. The package has been renamed to `rmf_traffic_editor_assets` hence it would fail and not spawn any thumbnails for models.

### Fix applied

Fixed the package name. The change will only apply to new users since the setting is saved in a hidden file in the home directory.
Existing users having the issue will need to either delete the settings file under (in Ubuntu it's under `~/.config/open-robotics/traffic-editor.conf`) or edit the field from preferences manually.


codebot (approved):
Thanks for catching and fixing this. :bow:

PR #344, opened 2021-04-26
Feature/layer rendering palette mapping

codebot:
### Implemented feature

The robot maps are now colorized and rendered in a nicer way to allow multiple robot occupancy-grid layers to be superimposed.

Upon reading a layer file, this implementation assumes that it's a robot-generated map using the palette style common to Cartographer or GMapping, with "mostly black" meaning a wall, "mostly white" meaning free space, and "somewhere in-between" meaning unknown. This PR maps those colors so that only the walls are strongly visible when rendered, using the color selected in the layer table. This lets you have different colors for the walls detected by each robot's map, which is helpful when sanity-checking robot map alignments and creating traffic lanes.

colorized_office_maps


Yadunund (approved):
Tested and works great ✨

Issue #343, opened 2021-04-23
Improving simulation rendering realism


luca-della-vedova:
### Description

It would be great to start using high resolution textures as well as more modern rendering workflows supported by Ignition (i.e. PBR or Physics Based Rendering) to increase realism.
Apart from the :sparkles: effect it's also important for people who would want a more realistic simulation environment for example for testing computer vision / localization algorithms in their robots together with RMF. 

### Implementation considerations

A current indoor environment (office) without advanced lighting and textures looks like this:

![image](https://user-images.githubusercontent.com/9111558/115856887-9c88ee00-a45f-11eb-9ba6-d70f4ec53cae.png)

While a [scenario](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Indoor%20lightmap) with lightmaps, high resolution textures, environment / roughness / normal maps looks like this:

![image](https://user-images.githubusercontent.com/9111558/115857174-f12c6900-a45f-11eb-8bdc-d20b6af16b28.png)

There is multiple steps to achieve such quality of rendering, all of them except n.1 are only applicable to Ignition since Gazebo classic doesn't support most modern rendering features:

- [ ] High resolution diffuse texture support #342 
- [ ] PBR textures support, (normal, roughness, metal maps) TODO
- [ ] Environment map generation and upload to fuel, reference the [ign-rendering tutorial](https://github.com/ignitionrobotics/ign-rendering/blob/main/tutorials/22_environment_map.md)
- [ ] Adding ceilings with textures for lights TODO
- [ ] Baking lightmaps and applying to the whole environment, [tutorial](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering5/tutorials/04_lightmap.md)

All of this should be done in a backward compatible way to avoid breaking behavior in Gazebo classic.

PR #345, opened 2021-04-26
PBR textures for walls and floors in Ignition Gazebo

luca-della-vedova:
### Implemented feature

This PR is a followup of #342, and is to work towards #343. It adds support to `rmf_building_map_tools` for PBR textures to be used in Ignition, specifically metalness, roughness, normal, environment and light maps.

The YAML format now has additional parameters for walls and floors that follow the same naming convention as the [SDF metal PBR workflow](http://sdformat.org/spec?ver=1.8&elem=material#pbr_metal).
For ease of review the PR doesn't yet include all the `rmf_traffic_editor` GUI work to include the parameters, hence they have to be added manually to the yaml file. An example gist for the office world that can be used to test the PR is [here](https://gist.github.com/luca-della-vedova/846103f59f0190dd3de59840052f49fe).

Followup work will be to add the parameters to the `rmf_traffic_editor` GUI and autopopulate them when possible (i.e. if a `[...]_Diffuse.png` texture is present, the Normal map should default to `[...]_Normal.png`)

Before the PR:
image

After the PR (notice the reflections on the floor for the ceiling lights given by the environment map. It looks better when moving around I promise!
image

While implementing I took the chance to do a small refactor of the wall parameters and how they were grouped by adding an equality operator to parameters, this made the implementation quite a bit cleaner (otherwise we had to keep comparing parameters manually every time a new one was added).

A small issue that I couldn't figure out yet is that the `model.material` file created [here](https://github.com/open-rmf/rmf_traffic_editor/compare/feature/pbr_textures?expand=1#diff-b13c13bd9ab2f46c4791fc58532ba225db2ad0a762a72efffa85325147ca8faaR74-R87) has duplicated entries which makes me think that  the generate function is called twice for each wall / floor.
It still works but it makes me wonder if there is some part of the map generator that is responsible for this issue and if there is a chance for optimizing it.

### DIFF
--- a/rmf_building_map_tools/building_map/floor.py
+++ b/rmf_building_map_tools/building_map/floor.py
@@ -292,7 +292,9 @@ class Floor:
             texture_name = self.params['texture_name'].value
 
         pbr_textures = get_pbr_textures(self.params)
-        texture_filename = add_pbr_material(visual_ele, model_name, f'floor_{floor_cnt}', texture_name, f'{model_path}/meshes', pbr_textures)
+        texture_filename = add_pbr_material(
+                visual_ele, model_name, f'floor_{floor_cnt}', texture_name,
+                f'{model_path}/meshes', pbr_textures)
 
         mtl_path = f'{model_path}/meshes/floor_{floor_cnt}.mtl'
         with open(mtl_path, 'w') as f:
--- a/rmf_building_map_tools/building_map/floor.py
+++ b/rmf_building_map_tools/building_map/floor.py
@@ -6,7 +6,7 @@ import shapely.ops
 
 from xml.etree.ElementTree import SubElement
 
-from .material_utils import copy_texture
+from .material_utils import add_pbr_material, copy_texture, get_pbr_textures
 from .param_value import ParamValue
 
 triangulation_debugging = False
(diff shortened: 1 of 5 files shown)


tanyouliang95:
Tested and works without breaking (Cant really tell the diff with a naked eye thou). One thing that I am unsure of, does the rendering works if we use the [.mtl with pbr field](http://exocortex.com/blog/extending_wavefront_mtl_to_support_pbr)?

luca-della-vedova:
> Tested and works without breaking (Cant really tell the diff with a naked eye thou). One thing that I am unsure of, does the rendering works if we use the [.mtl with pbr field](http://exocortex.com/blog/extending_wavefront_mtl_to_support_pbr)?

Not really for now sadly. I'm not sure why it is not implemented, one guess is that the mtl specification for PBR textures is a bit limited and doesn't offer support for all the maps we use (for example I don't think think there is a standard for `metalness` maps or `light` maps).

tanyouliang95 (approved):

luca-della-vedova:
With [ign-common #216](https://github.com/ignitionrobotics/ign-common/pull/216) PBR tags can now be added straight to the mtl, however there is still no way to add lightmaps since they usually require a different uv set to be mapped and the obj format doesn't support multiple uv sets.
Hence while the ign-common PR would help to get rid of the [sdf material override code](https://github.com/open-rmf/rmf_traffic_editor/pull/345/files#diff-b13c13bd9ab2f46c4791fc58532ba225db2ad0a762a72efffa85325147ca8faaR49-R93), it would make mapping light maps not possible. The code can probably be cleaned up in the future once we figure out how to use light maps in OBJ files.

PR #347, opened 2021-05-12
Feature/display layer transforms in freefleet format

codebot:
### Implemented feature

When clicking a layer in the layer-selection table, now the property editor will show the transformation parameters in a few different formats in the Property Editor.

This PR uses the transform currently being used to render a layer and computes it in a few different conventions. Because there are several different ways to parameterize transforms for different schemes (forward, backwards, translate-then-rotate, rotate-then-translate, etc.), it seems useful to automatically compute them and display more than one convention in `traffic-editor` so that users can copy/paste the data into the integration framework they are using. We have tested this for the current FreeFleet convention (which, in retrospect, isn't super intuitive...) in a few different maps.

The trick with this transform calculation is that most robot frameworks use the lower-left corner of the image as the origin, but in `traffic-editor` we are currently annotating in image space, where the upper-left corner of the image is the origin. This just requires a tiny bit of trig to convert between the spaces, which is implemented in this PR.

### DIFF
--- a/rmf_traffic_editor/gui/editor.cpp
+++ b/rmf_traffic_editor/gui/editor.cpp
@@ -98,6 +98,7 @@ Editor::Editor()
       {
         layer_idx = row;
         layer_table->update(building, level_idx, layer_idx);
+        populate_property_editor(building.levels[level_idx].layers[layer_idx-1]);
       }
     });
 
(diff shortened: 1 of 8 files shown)


Yadunund (approved):
Tested on hardware for `free_fleet` 👍🏼

PR #350, opened 2021-05-25
Fix/test maps build

luca-della-vedova:
### Fixed bug

When a map had an empty wall `texture_name` field (i.e. in `test_maps`), the default initialization assigned the value to a string while downstream code expected a ParamValue, this caused runtime errors.

### Fix applied

The assignment has been changed to ParamValue, also github actions have been updated to compile `test_maps` to test for this kind of regressions in the future.

Note to self / others, we should probably rename the package to `rmf_traffic_editor_test_maps` and not `test_maps`.


codebot (approved):
LGTM, thanks for chasing this down and for the clean fix to both the code and the CI

PR #352, opened 2021-05-31
Support non-symmetrical double swing doors

aaronchongth:
### Implemented feature

Supports non-symmetrical double swing doors. Screenshots below are examples after modifying door `n02_door` in the airport terminal demo.

image
Screenshot from 2021-05-31 19-04-44
Screenshot from 2021-05-31 19-05-20

This feature was implemented by adding a parameter `left_right_ratio` to door edges. With `1.0` as the default value.


codebot (approved):
nice! I tested it in the GUI and it looks great :sparkles:

PR #356, opened 2021-06-04
add world 'dae export ready' option, bugfixes for alpha values

ddengster:
## Bug fix/Fix applied

Alpha bugfixes
1) Diffuse/Specular color value standardization: In our sdf exporters we use 3 floats for rgb, this standardizes it to 4 values rgba with the alpha being 1.0. 
2) .mtl file alpha fixes: Python would output alpha values as <param.ParamValue....>, a fix has been applied to cast it properly to floating point. Before this fix walls would appear black when imported in blender after being exported to dae.

### Implemented feature

Add option "ignition_dae_export" to export world files ready for ignition to do dae exports.

This option enables our worlds to be 'dae export ready' via adding the correct plugin tags and removing any robots.

Related ignition PR: https://github.com/ignitionrobotics/ign-gazebo/pull/474
Downstream PR: https://github.com/open-rmf/rmf_demos/pull/53


rmf_building_map_tools/building_map/level.py
```diff
@@ -174,17 +174,18 @@ def generate_walls(self, model_ele, model_name, model_path):
...
                 self.elevation)
 
         # sniff around in our vertices and spawn robots if requested
-        for vertex_idx, vertex in enumerate(self.vertices):
-            if 'spawn_robot_type' in vertex.params:
-                self.generate_robot_at_vertex_idx(vertex_idx, world_ele)
+        if dae_export_plugin is False:
+            for vertex_idx, vertex in enumerate(self.vertices):
+                if 'spawn_robot_type' in vertex.params:
+                    self.generate_robot_at_vertex_idx(vertex_idx, world_ele)
```
luca-della-vedova, line 188: I wonder if we should limit the filtered objects to only robots or to all the dynamic objects, i.e. lifts / doors.
What would you recommend?

ddengster: yup i forgot about that, we should remove those

ddengster: removed ac3081a

Issue #361, opened 2021-06-21
Allow building_map_model_downloader to check export folder for models to determine if to download them even when using ignition fuel tools


nicholas-gs:
### Description

Currently, when using ignition fuel tools to download models (using the `-f` flag), `building_map_model_downloader` only checks `~/.ignition/fuel/fuel.ignitionrobotics.org` to see if the model is missing and needs to be downloaded.

However, a problem arises if using the `-e` flag as well. If a model is located in `~/.ignition/fuel/fuel.ignitionrobotics.org` but not in the folder specified using the `-e` flag, then the model will not be downloaded.

One option to solve this is to automatically check the specified export folder for missing models. And maybe if the model is already in `~/.ignition/fuel/fuel.ignitionrobotics.org` , we can simply export the model without having to download again.

Issue #360, opened 2021-06-21
building_map_model_downloader folders all in lowercase


nicholas-gs:
When using the `-f` flag to use ignition tools to download the models and specifying an export folder using the `-e` flag, the exported folders are all in lower case (following how ignition tools does it). But that results in gazebo being unable to find the models.

The solution is to have the export folder follow the capitalization of the original model names (as defined in the `.building.yaml` file).

PR #362, opened 2021-06-22
Fix issues with building map tools using Ignition tools

nicholas-gs:
### Fixed bug

1. #360. The folders downloaded by Ignition is always in lowercase, and so does the exported folders. This results in gazebo being unable to find the models.
2. #361 

### Fix applied
 
1. Commit 6fa825f. Have the exported folders follow the original name of the models as described in the `building.yaml` file.
2. Commit abe741b. If exporting models, automatically check the export path for missing models.

### DIFF
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -109,7 +109,11 @@ def download_models(
     # Ignition fuel tools can only download to this folder, so we set the
     # model path to it
     if fuel_tools:
-        model_path = "~/.ignition/fuel/"
+        if export_path is not None:
+            model_path = export_path
+            ign = False
+        else:
+            model_path = "~/.ignition/fuel/"
 
     missing_models = pit_crew.get_missing_models(
         model_set,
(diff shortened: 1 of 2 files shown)


rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
```diff
@@ -109,7 +109,11 @@ def download_models(
     # Ignition fuel tools can only download to this folder, so we set the
     # model path to it
     if fuel_tools:
-        model_path = "~/.ignition/fuel/"
+        if export_path is not None:
+            model_path = export_path
+            ign = False
```
aaronchongth, line 114: Wouldn't this revert back to using `http` to download the models after using both the `-f` and `-e` flags?

aaronchongth (changes requested):
Thanks for catching these bugs!  I just have a question below, but everything looks pretty good.

rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
nicholas-gs: Hello Aaron. No it would not. The `ign=False` is just so that the it can search the folder specified by `-e` for missing models. The models will still be downloaded using the new Ignition Fuel tools. I've tested it out in this [PR](https://github.com/open-rmf/rmf_demos/pull/57) for `rmf_demos` to check that its the case.

aaronchongth: Thanks for explaining, didn't catch these lines, https://github.com/nicholas-gs/rmf_traffic_editor/blob/fix/building_map_tools/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py#L154-L157

aaronchongth (approved):
LGTM! Thanks for fixing it so quickly!

PR #364, opened 2021-06-25
Minor tweak to how empty crowd_sim and lift structures are serialized in YAML

codebot:
This is a trivial PR that explicitly sets how whitespace (carriage returns and indents) are handled for a few places when serializing the `crowd_sim` data structure in YAML. The two forms are equivalent; this just causes `yaml-cpp` to generate whitespace in the same way that the JavaScript `yaml` library generates it when the emitter comes upon an empty sequence nested inside a block-style map. Otherwise, the two implementations (C++ and JavaScript) would generate whitespace diffs each time they saved each other's files.

Similarly, for empty lift maps (i.e. building models that do not have lifts), this PR forces a flow-style empty map `{}` to be generated in `yaml-cpp`, to force the whitespace to be generated identically by the `yaml-cpp` and JavaScript `yaml` libraries.

### DIFF
--- a/rmf_traffic_editor/gui/building.cpp
+++ b/rmf_traffic_editor/gui/building.cpp
@@ -164,6 +164,8 @@ bool Building::save()
   y["lifts"] = YAML::Node(YAML::NodeType::Map);
   for (const auto& lift : lifts)
     y["lifts"][lift.name] = lift.to_yaml();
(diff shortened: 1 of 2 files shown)


Yadunund (approved):
Looks good 😄

PR #372, opened 2021-07-28
Feature: align vertices colinear

codebot:
### Implemented feature

This PR adds multiple vertex selection by holding down SHIFT and clicking vertices to add them to the current selection. All selected vertices will all be rendered as red. You can clear the current selection by pressing the Escape key, as before.

When multiple vertices are selected that are directly connected by edges, you can press the slash key `/` or use the new menu item `Edit->Align Colinear`, and it will attempt to compute an orthogonal projection of all intermediate vertices onto the line between the first and last vertices in the selected chain.

There are probably corner cases that are not handled correctly, but at least it seems to work in the few situations that I've tried so far.

### DIFF
--- a/rmf_traffic_editor/gui/building.h
+++ b/rmf_traffic_editor/gui/building.h
@@ -172,12 +172,14 @@ public:
     std::vector<EditorModel>& editor_models,
     const RenderingOptions& rendering_options);
 
+  /*
   void mouse_select_press(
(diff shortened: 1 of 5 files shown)


Yadunund (approved):
Thank you so much for adding this so quickly! Works great after the latest changes 💯

PR #373, opened 2021-07-30
sort list by model name, not Fuel group name

codebot:
This PR sorts the "add model" dialog box by the model name, instead of by the entire group_name/model_name string. This makes typing names quickly into the dialog edit box much faster. For example, as you start typing "desk", it will scroll the list to `OpenRobotics/Desk` and show the relevant preview. You don't have to type `OpenRobotics/Desk`


Yadunund (approved):
This is super helpful! Thank you 🙇🏼

PR #374, opened 2021-08-04
resolve build error on some systems with size_t namespace

codebot:
On some systems and some compilers, apparently the cascade of includes in some files in traffic-editor does not include `size_t` outside the `std` namespace, for some files. This PR just adds `std::` in front of all uses of `size_t` so that the code is more portable.


Yadunund (approved):
👍🏼

Yadunund (approved):

PR #377, opened 2021-08-11
Feature/wall graph

mattbooker:
### Implemented feature

The `building_map_server` now populates the wall graph field within the Level message


rmf_building_map_tools/building_map/level.py
```diff
@@ -487,3 +487,46 @@ def add_lanes_from(self, other):
...
+        for w in self.walls:
+            v1 = self.vertices[w.start_idx]
+            v2 = self.vertices[w.end_idx]
+
+            start_idx = vidx_to_mapped_idx[w.start_idx]
+            end_idx = vidx_to_mapped_idx[w.end_idx]
+
+            wall_data['walls'].append([start_idx, end_idx, w.params])
+            
+        return wall_data
```
arjo129, line 532: Knit: We always end files with a newline. 
See: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

arjo129 (changes requested):
These changes could be useful outside of the current ar-kit context as well. There are several minor code style changes needed as marked by the `pycodestyle` checker. 
https://github.com/open-rmf/rmf_traffic_editor/pull/377/checks?check_run_id=3298220632

Also all files should end in a new line.

Issue #375, opened 2021-08-04
Crowd simulation map generation fails when no robots are in the map


luca-della-vedova:
### Required information:

Build from source, current main branch on Foxy and Ubuntu 20.04

### Description of the bug

When there are no robots in the map the agent group `external_agent`, used to track robots, has 0 agents and this causes a failed assertion when the AgentGroup is checked for validity [here](https://github.com/open-rmf/rmf_traffic_editor/blob/main/rmf_building_map_tools/building_crowdsim/config/scene_file.py#L158-L159)

A possible fix could be to skip the group instead of having a failed assertion if it has no agents, or to have such condition only for the `external_agent` group that includes robots.

### Steps to reproduce the bug

Create a map that has no robots but has crowdsim enabled, then try to build the workspace.

luca-della-vedova:
Fixed by rmf_traffic_editor [#380](https://github.com/open-rmf/rmf_traffic_editor/pull/380).

PR #380, opened 2021-08-12
fix crowdsim map generation when there are no robots

xiyuoh:
### Fixed bug

Fixed bug #375: `building_crowdsim.py` unable to generate the required files for crowd simulation when no robot is added to the map due to a `ValueError` where the AgentGroup element fails its validity check

### Fix applied

Added a condition in `configfile_generator.py` to only append AgentGroup to the scene file when there are robots in the map, denoted by length of the AgentGroup sub-elements.

Code can be tested by running `building_crowdsim.py` with a `.building.yaml` file with no robots added

### DIFF
--- a/rmf_building_map_tools/building_crowdsim/config/configfile_generator.py
+++ b/rmf_building_map_tools/building_crowdsim/config/configfile_generator.py
@@ -73,7 +73,9 @@ class ConfigFileGenerator:
             for item in self.crowd_sim_yaml['agent_groups']:
                 cur_group = AgentGroup()
                 cur_group.load_from_yaml(item)
-                self.scene_file.sub_elements.append(cur_group)
+                cur_generator = cur_group.sub_elements[-1]
+                if len(cur_generator.sub_elements) != 0:
+                    self.scene_file.sub_elements.append(cur_group)
 
         write_xml_file(
             self.scene_file.output_xml_element(),
--- a/rmf_building_map_tools/building_crowdsim/config/configfile_generator.py
+++ b/rmf_building_map_tools/building_crowdsim/config/configfile_generator.py
@@ -73,8 +73,7 @@ class ConfigFileGenerator:
             for item in self.crowd_sim_yaml['agent_groups']:
                 cur_group = AgentGroup()
                 cur_group.load_from_yaml(item)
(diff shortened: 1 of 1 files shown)


youliangtan:
@xiyuoh Thks for the PR. One thing here, you will need to re-commit all your commits (force push) with [GPG](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits) signed, and also a sign-off. These can be done by setting up gpg, then `git commit -s -S` (note that `-s` and `-S` are different)

xiyuoh:
@youliangtan sorry about that, thanks for letting me know! I think i re-committed successfully, lmk if there are other issues :)

youliangtan:
I noticed that only 1 commit is "signed". Unfortunately all past commits need to get signed. One way is to force remove all those commits and create a new one.

xiyuoh:
@youliangtan hi! i removed the unverified commits, hopefully its good now?

rmf_building_map_tools/building_crowdsim/config/configfile_generator.py
```diff
@@ -73,7 +73,9 @@ def generate_scene_file(self, output_dir):
             for item in self.crowd_sim_yaml['agent_groups']:
                 cur_group = AgentGroup()
                 cur_group.load_from_yaml(item)
-                self.scene_file.sub_elements.append(cur_group)
+                cur_generator = cur_group.sub_elements[-1]
+                if len(cur_generator.sub_elements) != 0:
```
youliangtan, line 77: After a quick look. is it possible for us to use [this](https://github.com/open-rmf/rmf_traffic_editor/blob/8444941cf1633747eab49e0b2cbd9cd8d299627f/rmf_building_map_tools/building_crowdsim/config/scene_file.py#L157) `AgentGenerator.is_valid()` function?

youliangtan (changes requested):

rmf_building_map_tools/building_crowdsim/config/configfile_generator.py
xiyuoh: good idea! thanks for the suggestion

youliangtan (approved):
LGTM :+1:

PR #382, opened 2021-08-17
added scaling features to wall texture

xiyuoh:
Signed-off-by: xiyuoh <ohxiyu@gmail.com>

### Implemented feature

This PR follows #343 to add scalability to downloaded textures from URL. 

As mentioned in #342 the custom textures may be stretched in simulation as their height-to-width ratio may not be coherent with the wall dimensions (walls expect a 2.5m x 1m texture size). This feature implemented allows users to customize texture scales by adding parameters to define these scales.

Before implementation:
Screenshot from 2021-08-17 16-31-04 (1)
You can observe the texture stretching by comparing the wall and floor textures.

After implementation:
Screenshot from 2021-08-17 16-41-52 (1)

The scaling is controlled by three parameters: `texture_height`, `texture_width` and `texture_scale`:
SS2 (1)

- `texture_height` and `texture_width`: dimensions of the texture image. It does not need to be absolute as only their ratio will be calculated, i.e. for square textures, you can enter 1 for both parameters.
  But for images where the ratio may not be easy to compute, users can enter the absolute dimensions directly.
- `texture_scale`: how 'zoomed in' you want your texture to be (default is 1)
   For example, if you want your texture width to be 1m, enter 1. If you want your texture width to be 2m, enter 2. The height will be scaled accordingly.
   Special case: If you want your texture image to fit the full wall (texture height = wall height), enter 0.

See some examples below!

For `texture_scale = 2` with square textures
Screenshot from 2021-08-17 16-30-28 (1)
SS3 (1)

For `texture_scale = 0` with complicated texture dimensions
Screenshot from 2021-08-17 16-31-11 (1)
SS1 (1)

### Testing the PR

1. Open a map in `traffic_editor`
2. Add texture URLs and scaling parameters like the examples above
3. Launch the demo, or run `building_map_generator.py gazebo` on your yaml map and open the mesh files in gazebo.

### DIFF
--- a/rmf_building_map_tools/building_map/wall.py
+++ b/rmf_building_map_tools/building_map/wall.py
@@ -21,6 +21,9 @@ class Wall:
         self.texture_name = wall_params['texture_name']
         self.alpha = wall_params['alpha']  # val 0.0-1.0 transparency of wall
         self.pbr_textures = get_pbr_textures(wall_params)
+        self.texture_height = wall_params['texture_height']
+        self.texture_width = wall_params['texture_width']
+        self.texture_scale = wall_params['texture_scale']
 
         # Wall filtering according to wall_params
         for checked_wall in yaml_node:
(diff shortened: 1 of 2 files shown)


codebot (changes requested):
This is great! Thank you for the clean diff and the great illustrations.

Because there are many `.building.yaml` files around that people will want to keep using without having to re-open and save them in an updated traffic-editor GUI, can you add checks for the `texture_width` and other new parameters in the `Wall` class, so that it won't crash when opening an "old" `.building.yaml` file? You could test it against the existing demo maps that don't have these "new" parameters. Thanks!

xiyuoh:
@codebot done, thanks for the feedback!

PR #383, opened 2021-08-20
added support for ceilings with texture

xiyuoh:
Signed-off-by: xiyuoh <ohxiyu@gmail.com>

### Implemented feature

This PR follows #343 to add ceilings with texture to simulation environments. The ceiling is generated by writing an additional OBJ file to the meshes. 

Additional floor parameters are now available in the traffic-editor GUI for users to customise their environments:
- `indoor`: enter 1 or 0 to indicate whether the environment is indoors. If 1, a polygon with the same shape as the floor space will be added.
- `ceiling_texture`: applies texture to ceiling from URLs
- `ceiling_scale`: applies scaling to ceiling texture

Implemented ceilings are only visible from within the indoor space; from the bird-eye view, users can still see the interior of the environment without the ceilings blocking.

Example:
Screenshot from 2021-08-20 12-07-09
Screenshot from 2021-08-20 14-10-07

### DIFF
--- a/rmf_building_map_tools/building_map/floor.py
+++ b/rmf_building_map_tools/building_map/floor.py
@@ -21,6 +21,7 @@ class Floor:
         self.vertex_indices = []
         self.vertices = []
         self.thickness = 0.1
+        self.wall_height = 2.5
         if 'vertices' in yaml_node and yaml_node['vertices'] is not None:
             self.vertex_indices = [i for i in yaml_node['vertices']]
 
@@ -28,6 +29,10 @@ class Floor:
         if 'parameters' in yaml_node and yaml_node['parameters']:
             for param_name, param_yaml in yaml_node['parameters'].items():
                 self.params[param_name] = ParamValue(param_yaml)
+
+        self.indoor = 0
+        if 'indoor' in self.params:
+            self.indoor = self.params['indoor'].value
 
     def to_yaml(self):
         y = {}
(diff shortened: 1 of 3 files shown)


luca-della-vedova:
Thanks for the PR! I think we would want to support different set of PBR textures for the floor and its matching ceiling, for example a ceiling might have an `emissive` map (to model lights) while the floor wouldn't, and maybe the floor would have a `roughness` map while the ceiling wouldn't (for example for a concrete floor).
From a quick glance it seems that this use case is not supported, I think it's OK not to have it in the `traffic_editor` GUI but it should be supported at least at the yaml level. An example of how PBR textures are implemented currently for floors is documented in #345 while the example yaml is [here](https://github.com/login/oauth/authorize?client_id=7e0a3cd836d3e544dbd9&redirect_uri=https%3A%2F%2Fgist.github.com%2Fauth%2Fgithub%2Fcallback%3Freturn_to%3Dhttps%253A%252F%252Fgist.github.com%252Fluca-della-vedova%252F846103f59f0190dd3de59840052f49fe&response_type=code&state=01e15a656a67d2c895a67fe243b141b711d2c7875d44525cb5efc392e637772c).
A possible implementation would be to have parameters such as `ceiling_roughness_map`, `ceiling_emissive_map` etc.

xiyuoh:
@luca-della-vedova Ok! Added support for separate PBR textures. I'm able to generate the model with these textures successfully, but can't really seem to see the difference (even the ceiling lights). The changes are reflected in the [model SDF](https://gist.github.com/xiyuoh/602de3645306d26ead4a038cc7f6bfcd#file-model-sdf-L55-L59), so I'm not sure if it's just not too visible or if it isn't working. 
On another note, can I check if you were able to open your example yaml file in the `traffic-editor` GUI? I have been getting a bad conversion error when using yaml files with the PBR parameters.

Issue #386, opened 2021-09-06
Certain fuel models seem missing


cnboonhan:
I encountered some issues with running airport_terminal under ignition gazebo, where some models ( such as Construction Cone ) could not be found after a colcon build, and this prevented ignition gazebo from fully launching. 

and so I tried to debug by running this command standalone:

```
 ros2 run rmf_building_map_tools building_map_model_downloader install/rmf_demos_maps/share/rmf_demos_maps/airport_terminal/airport_terminal.building.yaml -f -e install/rmf_demos_maps/share/rmf_demos_maps/airport_terminal/maps/airport_terminal_ign/models`
```

For certain models, the following error appears:
```
ERROR::pit_crew.download_model_fuel_tools():702: Could not download model 'Construction Cone'! [Errno 39] Directory not empty: '/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/construction%20cone' -> '/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/construction cone'

```

Models that are successful seem to explicitly show that a model is exported to the target folder:
```
Download succeeded.
WARNING::pit_crew.sync_sdf():924: Synced SDF names for kitchensink! Changed from KitchenSink to kitchensink
WARNING::pit_crew.sync_sdf():967: Synced SDF names for kitchensink! Changed from KitchenSink to kitchensink
Exporting KitchenSink to install/rmf_demos_maps/share/rmf_demos_maps/maps/airport_terminal_ign/models/KitchenSink
```

After all this, it does seems that some models, like "Construction Cone", are not found in the correct folders:
```
root@rmf:/opt/rmf# ls install/rmf_demos_maps/share/rmf_demos_maps/maps/airport_terminal_ign/models/
Bookshelf  Chair                 FemaleVisitor     Fountain  Gazebo             KitchenSink   MainTable         MaleVisitorSit  MopCart3           Playground       SUV          Table          WhiteChipChair
Cafe       ElectronicsRecycling  FemaleVisitorSit  Fridge    KitchenCountertop  LampAndStand  MaleVisitorPhone  MiniSofa        ParkingTrolleyMax  PotatoChipChair  StorageRack  VisitorKidSit  airport_terminal_L1
root@rmf:/opt/rmf# ls install/rmf_demos_maps/share/rmf_demos_maps/maps/airport_terminal
airport_terminal/     airport_terminal_ign/
[...]

PR #387, opened 2021-09-08
fix whitespace issue in download chain

cnboonhan:
Addresses https://github.com/open-rmf/rmf_traffic_editor/issues/386

Current implementation:

* downloads model to ~/.ignition folder, with whitespaces in names parsed as "%20"
* checks if there is whitespace in the model name
* If whitespace exists, rename "%20"-formatted name to whitespace name

the rename step fails when a model already exists with the whitespace name, which happens if the model was already downloaded in another demos package. This affects the export step involving the copy of the model from ~/.ignition folder to the workspace install folder.

PR is a quick fix targeting the rename step by ensuring any pre-existing model folder is removed, thus avoiding this failure. In the long term, it might be worth stripping whitespaces in fuel itself?

### DIFF
--- a/rmf_building_map_tools/pit_crew/pit_crew.py
+++ b/rmf_building_map_tools/pit_crew/pit_crew.py
@@ -638,6 +638,8 @@ def download_model_fuel_tools(model_name, author_name,
             model_name.lower())
 
         if " " in model_name:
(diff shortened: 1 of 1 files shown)


aaronchongth (approved):
Testing and it's working well! Thanks!

PR #389, opened 2021-09-15
Add support for Cartesian Meters reference system and coordinate offset

luca-della-vedova:
### Implemented feature

This PR (mainly done by @codebot ) introduces a new `CoordinateSystem` object in `rmf_building_map_tools` that is mainly used to implement a Cartesian System with matching x-y coordinate conventions (where the Y axis is flipped compared to the usual image coordinates).
It also introduces an offset field that can be set in projects to report to downstream packages that all the coordinates should be offset by a certain constant.
This is especially useful when cartesian systems in global coordinates, such as [EPGS:3414/SVY21](https://epsg.io/3414) are used, since they can have large offsets.

The offset field is currently not published by `building_map_server` because there is no support in the top level [BuildingMap message](https://github.com/open-rmf/rmf_building_map_msgs/blob/main/rmf_building_map_msgs/msg/BuildingMap.msg) for parameter fields.
A new message is being worked on, together with a matching publisher in [rmf_building_map_msgs #3](https://github.com/open-rmf/rmf_building_map_msgs/pull/3) and the [feature/sitemap_server branch](https://github.com/open-rmf/rmf_traffic_editor/tree/feature/sitemap_server) in this repo to expose information on the coordinate system to downstream users.

### DIFF
--- a/rmf_building_map_tools/building_crowdsim/building_yaml_parse.py
+++ b/rmf_building_map_tools/building_crowdsim/building_yaml_parse.py
@@ -4,12 +4,13 @@ import os
 import copy
 
 from building_map.building import Building
+from building_map.coordinate_system import CoordinateSystem
 from building_map.level import Level
 
 
 class LevelWithHumanLanes (Level):
-    def __init__(self, yaml_node, name, graph_idx=9):
-        Level.__init__(self, yaml_node, name)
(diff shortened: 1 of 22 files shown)


luca-della-vedova:
Superseded by #396

Issue #388, opened 2021-09-10
Deprecate option to use http request for model downloading


aaronchongth:
### Deprecation

Related to https://github.com/open-rmf/rmf_traffic_editor/issues/386.

We should use `ign-fuel-tools` moving forwards, and support upstream development if we need any new features.

cnboonhan:
happy to work on this, but I would like to get a clear idea of the ideal result. From [here](https://github.com/open-rmf/rmf_traffic_editor/issues/386#issuecomment-914999932): 
```
A real fix would be to throw all the models, including the building and floor generated by rmf_building_map_tools into a common folder, to also remove the current mesh duplication.
```

I am guessing, we the model download process should be something like
* download from fuel to `~/.ignition`, ignoring duplicates
* export models to dedicated `install` folder, perhaps `install/rmf_fuel_models`
* Link all launch files to this folder 

Some thoughts I had:
* Ignition seems to halt if some models are missing. This might not be ideal for CI, which might want to omit some models for the sake of efficiency. Is there a way to let ignition continue running with some missing models? 
* Would there be value in improving the old models we have to remove whitespace, etc? @aaronchongth  mentioned some stuff about some standards to enforce on fuel models, which we could spend some time to tidy up if ideal.

aaronchongth:
As per VC, this issue will only be related to refactoring and deprecating the option to download models using the `http` request feature.

Regarding, the discussion about exporting models, I have created a separate issue https://github.com/open-rmf/rmf_demos/issues/89.

As for reducing the number of workarounds we apply during model downloading, e.g. handling white space, changing names, etc, as you surmised, the ideal solution would be to improve the older models on Fuel, however I would expect plenty of users are using them as it is, and any major changes would break lots of simulation behaviors across the board. We should definitely revisit this once we get a better understanding on our simulation model needs some time soon. Thanks!

cnboonhan:
yes thanks for clarifying! will dive in 🤽

PR #395, opened 2021-09-27
Deprecate/http download

cnboonhan:
## New feature implementation
This PR aims to deprecate http downloads according to [this PR](https://github.com/open-rmf/rmf_traffic_editor/issues/388).

### Implemented feature

- Remove all logic related to htttp model downloads
- print a helpful message for users to remove the -f flag from their builds

I switched out the `-f` option with a helpful printout to prevent builds from failing. However, `colcon` seems to hide this printout, even though i print to `stderr`. Ideally we should perhaps make this change very visible during `colcon` builds.


rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
```diff
@@ -166,12 +148,13 @@ def download_models(
 
 def main():
     args = parser.parse_args()
+    print("The -f option of package building_map_mode_downloader has been deprecated. \
```
aaronchongth, line 151: This gets printed regardless if the flag `-f` was used, I would suggest a cleaner deprecation using `argparse.Action`, I found an example gist here, https://gist.github.com/alzix/4868d66115a55567d3b43b32742f6f4f

rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
```diff
@@ -101,27 +98,16 @@ def download_models(
...
-    logging.info("\nUsing Ignition Fuel directory struture : %s\n" % (ign))
     # Ignition fuel tools can only download to this folder, so we set the
     # model path to it
-    if fuel_tools:
-        if export_path is not None:
-            model_path = export_path
-            ign = False
-        else:
-            model_path = "~/.ignition/fuel/"
+    model_path = "~/.ignition/fuel/"
```
aaronchongth, line 103: We should probably start module/function level variables, according to https://google.github.io/styleguide/pyguide.html, we can probably do something like 
```
IGN_FUEL_MODEL_PATH = "~/.ignition/fuel/"
model_path = IGN_FUEL_MODEL_PATH
```

This way, we can probably remove the 2 lines of comments above it too.

aaronchongth (changes requested):
Thanks for handling this deprecation, I've added in a few comments, let me know what you think!

rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
cnboonhan: Addressed in e6744abdc2e7cafa00d144f9ad72d415817e64c1. I extended the deprecation notice to the -m flag as well, which is redundant now.

aaronchongth (approved):
LGTM!

PR #398, opened 2021-10-19
minor usability enhancements on traffic editor

youliangtan:
No major feature is being introduced here. Just made some cleanups `traffic-editor` while building a new rmf world, mainly for usability. Listed down some of the changes here: 

- default "add_lane" as bidirectional, as bidir is commonly used
- have a default string for "door_name" to prevent built rmf gazebo world from crashing due to empty str
- added more shortcut key ("D" for add door), and display it all valid shortcuts with tool tip
- ~~When adding Lift, a vertex will generate automatically on each floor. In `vertex: [param: { "lift_cabin"}]`, I noticed thtat the `lift_cabin` is not being used by any downstream pkgs in rmf. As tested, as long as a vertex is being created in the lift cabin, the robot can transverse through floors. Remove this unnecessary param to prevent confusion. Will update this in the doc.~~ Prevously, lift vertex with auto generated lift_cabin param can be deleted accidentally (which what i did). Make it not delete-able by user.
- Added some new assets (created previously) to `rmf_traffic_editor_assets`: `InformationCounter`, `Escalator`, `RetailKiosk`...
- Fix wall model transparency when changing alpha value of the wall. 
- Update Readme
- Instead of a blank icon, created and added a new window thumbnail icon for `traffic-editor`. A png version is attached here:

(srry for combining multiple small pieces of not so relevant changes into these pr)

### DIFF
--- a/rmf_building_map_tools/building_map/wall.py
+++ b/rmf_building_map_tools/building_map/wall.py
@@ -241,7 +241,7 @@ class Wall:
 
         visual_ele = SubElement(link_ele, 'visual')
         visual_ele.set('name', f'wall_{self.wall_cnt}')
-        
+
         # Note: Material visual profile in the sdf will override the profile
         # in .mtl in the obj file. TODO will need to soon allign both profiles
         v_geom_ele = SubElement(visual_ele, 'geometry')


rmf_traffic_editor/gui/lift_dialog.cpp
```diff
@@ -417,47 +412,6 @@ void LiftDialog::ok_button_clicked()
...
-        if ((it != v.params.end()) && (it->second.value_string == _lift.name))
-        {
-          v.x = to_point.x();
-          v.y = to_point.y();
-          found = true;
-        }
-      }
-      if (!found)
-      {
-        _building.add_vertex(level_idx, to_point.x(), to_point.y());
```
codebot, line 452: I agree that no downstream code uses the `lift_cabin` parameter directly (at least, I couldn't find any with `grep`), I think this code block is still required in order to run this line, which adds (or modifies, in the previous clause) the vertex for the center of the lift on each floor that the lift services. I thought this vertex was necessary in order to connect the lift cabin to the lane graph? If we delete this entire block, then the lift cabins won't have a vertex created for them, and as you move the lift around in the GUI, you'd have to manually adjust the lift center vertex. I guess that's OK, but this just seems convenient.

youliangtan: You are right. Currently, the only usage of "lift_cabin" is to move/create the vertex accordingly when the lift changes. I am also quite unsure about deleting this. 
For some context, the user can directly create 1 (or more) waypoint within the lift cabin, and building_map_tool will populate the graph.yaml file with ` vertices: - {lift: Lift1, name: ''}`. Initially, I wrongly thought that a "lift_cabin" param is needed.  Anyway, retaining this feature will also work flawlessly.

codebot: Yeah I think everything else in this PR is super great! But I do think we should leave (at least for now) this automatic cabin-vertex creation and updating block.

PR #401, opened 2021-11-14
Improve behavior for Cartesian maps

codebot:
Because Cartesian maps have 1-meter units, we need to compute
a better default scale for them. Previously it was always using
a default scale of 0.05.

Somewhat related, also fix the edge-select implementation so it
spins through all edges in the map rather than just selecting
the first edge within 10 units... that was OK when we were always
using pixel-based maps, but now that meter-based maps are in use,
it was just choosing the first edge that was within 10 meters of
the click, which was a lot of edges and felt somewhat random.


luca-della-vedova (approved):
Works great, thanks for the fix!

PR #405, opened 2021-12-03
flip layer images right-side up in cartesian meters mode

codebot:
Previously, when editing maps with `cartesian_meters` coordinate system, layer images would render upside-down. This PR fixes that. No change to maps with `reference_image` coordinate system. Tested on all demo maps (which are all `reference_image` based) and they all render as expected.


cnboonhan (approved):
This fixes it! thank you :bow:

PR #408, opened 2021-12-17
Handle simulation offsets for models and cameras

cnboonhan:
Hoping this patch will help handle the 

- latlon translation behavior for models ( previously, builds break when adding models due to WGS84 translation not having a "rotation" variable
- change naming convention for model positions to lat/lon instead of x/y for consistency
- apply offsets to the camera as well, so the camera view appears immediately over the working area

### DIFF
--- a/rmf_building_map_tools/building_map/building.py
+++ b/rmf_building_map_tools/building_map/building.py
@@ -454,7 +454,8 @@ class Building:
 
         gui_ele = world.find('gui')
         c = self.center()
-        camera_pose = f'{c[0]} {c[1]-20} 10 0 0.6 1.57'
+        # Transforming camera to account for offsets
+        camera_pose = f'{c[0] - self.global_transform.x} {c[1]-20 - self.global_transform.y} 10 0 0.6 1.57'
         # add floor-toggle GUI plugin parameters
         if 'gazebo' in options:
             camera_pose_ele = gui_ele.find('camera').find('pose')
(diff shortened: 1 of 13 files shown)


codebot (changes requested):
I don't think we always want to call the model coordinates `lat` and `lon` because there are many maps which are still Cartesian, and for maps which will remain in "globally defined but still cartesian planes, not lat/lon", such as SVY21, the `(x, y)` description is still valid. Many GIS libraries deal with this by just calling them "coordinates" instead of x/y or lat/lon. This leads, of course, to the ever present confusion about if `coords[0]` is latitude or longitude... we'd just want to do our own convention, probably following the ordering of GeoJSON.

But I'd suggest actually just keeping it `x, y` by passing and applying the world transform in the `Model` constructor. It could be passed in conveniently from where the `Model` instances are created in `level.py`. That way the whole lat/lon conversion is over during construction, and the rest of the code can be unchanged.

In order to keep the `to_yaml()` working, for now I'd suggest saving the "original" world coordinates in the constructor and just dumping them out again, unmodified, in `to_yaml()`. Long-term we could implement the inverse transform in our `Transform` class to do this properly, but we don't need that immediately. I don't think we use `to_yaml()` anymore.

cnboonhan:
Hopefully I understood correctly for these changes
* revert to using xy conventions: https://github.com/open-rmf/rmf_traffic_editor/pull/408/commits/2051ef7665aec9dbe180ece4257036f64ff337ea
* Pass the transform in constructor instead of `generate()`: https://github.com/open-rmf/rmf_traffic_editor/pull/408/commits/fd9a51de03ec4640dfc2f652da78352247166b57
* Add a check for existence of `global_transform` before using it to offset camera: https://github.com/open-rmf/rmf_traffic_editor/pull/408/commits/3c2f3e73b63546eead308f542837f9682949faca

codebot (approved):

PR #412, opened 2022-01-03
Always use ign=True and remove spaces when comparing model names

mxgrey:
This PR addresses two issues related to fetching models from fuel:

* We will now always use the `ign=True` mode since we are always caching models in the `.ignition/fuel` model cache.
* We will ignore all spaces in model names when deciding if a local model copy matches a model on fuel. Sometimes the models on fuel are inconsistent about their use of spaces, and when that happens pit crew would accidentally re-download it (but this is fixed in this PR).

### DIFF
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -7,6 +7,7 @@ from pprint import pprint
 import argparse
 import logging
 import sys
+import shutil
 import os
 import yaml
 
@@ -77,6 +78,14 @@ def get_crowdsim_models(input_filename):
             logger.error(f"Could not get crowd_sim models, error: {e}."
                          " Ignore models in crowd_sim...")
         return actor_names
+
+
+def export_downloaded_model(model_path, export_path):
+    if os.path.exists(export_path) or os.path.isdir(export_path):
+        logger.warning("%s already exists, skipping..." % (export_path))
+        return
+    logger.info("Exporting to %s..." % (export_path))
+    shutil.copytree(src=model_path, dst=export_path)
 
 
 def download_models(
@@ -153,6 +162,23 @@ def download_models(
     if missing_models.get('missing', []):
         logger.warning("\nMissing models (not in local or Fuel):")
         pprint(missing_models['missing'])
+
+    if export_path is not None:
+        available_models = missing_models.get('available', [])
+        for model_name, paths in available_models:
+            if paths is None:
+                logger.warning("No path for exporting model %s, skipping..."
+                               % (model_name))
+                continue
+
+            if len(paths) > 1:
+                logger.warning("More than one path for model %s, selecting "
+                               "first path that was found, %s"
+                               % (model_name, paths[0]))
+
+            export_model_path = os.path.join(export_path, model_name)
+            export_model_path = os.path.expanduser(export_model_path)
+            export_downloaded_model(paths[0], export_model_path)
 
 
 def main():
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -125,7 +125,8 @@ def download_models(
         cache_file_path=cache,
         lower=True,
         priority_dir=include,
-        ign=True
+        ign=True,
+        use_dir_as_name=True
     )
 
     logger.info("\n== REQUESTED MODEL REPORT ==")
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -51,7 +51,7 @@ parser.add_argument("-c", "--cache", type=str,
 parser.add_argument("-f", "--fuel-tools", action=HTTPDownloadDeprecated,
                     help="Use ignition fuel tools to download models instead "
                          "of http", nargs=0)
-parser.add_argument("-i", "--include", type=str,
+parser.add_argument("-i", "--include", type=str, default=None,
                     help="Search this directory first for models.")
 parser.add_argument("-e", "--export-path", type=str, default=None,
                     help="Export model downloaded using ignition fuel tools "
@@ -116,6 +116,7 @@ def download_models(
         cache_file_path=cache,
         lower=True,
         priority_dir=include,
+        ign=True
     )
 
     logger.info("\n== REQUESTED MODEL REPORT ==")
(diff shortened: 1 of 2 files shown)


aaronchongth:
Thanks for making the fix so quickly! After some testing, I found that this fix uncovered a bug that was hidden away due to `fuel_tools` always re-downloading the models. All the available models are not exported to the `install` directory when called `download_models` is called.

All the demo worlds should start complaining that models are missing or can't be found when launched.

```
rm -rf build/rmf_demos_maps install/rmf_demos_maps
colcon build --packages-select rmf_demos_maps
source install/setup.bash
ros2 launch rmf_demos_gz office.launch.xml
ros2 launch rmf_demos_ign office.launch.xml
```

methylDragon:
> Thanks for making the fix so quickly! After some testing, I found that this fix uncovered a bug that was hidden away due to `fuel_tools` always re-downloading the models. All the available models are not exported to the `install` directory when called `download_models` is called.
> 
> All the demo worlds should start complaining that models are missing or can't be found when launched.
> 
> ```
> rm -rf build/rmf_demos_maps install/rmf_demos_maps
> colcon build --packages-select rmf_demos_maps
> source install/setup.bash
> ros2 launch rmf_demos_gz office.launch.xml
> ros2 launch rmf_demos_ign office.launch.xml
> ```

Would recommend editing the downloader script for that, I don't think pit_crew should be responsible for handling exports in the download function since it's one level up from what pit_crew is supposed to be concerned with.

aaronchongth:
@methylDragon, yup! https://github.com/open-rmf/rmf_traffic_editor/pull/414 just allows `pit_crew` to return the paths, while the exporting happens on the downloader

aaronchongth:
Yeah I played around with the code trying not to break any other uses of `pit_crew`'s API :bow:

aaronchongth:
After testing the PR and fixing some minor issues,

With no pre-downloaded ignition fuel models, building `rmf_demos_maps` takes 4 minutes+,
```bash
rm -rf ~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models
mkdir ~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models

rm -rf build/rmf_demos_maps install/rmf_demos_maps
colcon build --packages-select rmf_demos_maps

# Check that models have been downloaded using fuel
ls ~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models

# Check that models have been exported to install
ls install/rmf_demos_maps/share/rmf_demos_maps/maps/DEMO/models
```

After that, re-building a fresh `rmf_demos_maps` takes around 16 seconds,
```bash
rm -rf build/rmf_demos_maps install/rmf_demos_maps
colcon build --packages-select rmf_demos_maps

# Check that models have been exported to install
ls install/rmf_demos_maps/share/rmf_demos_maps/maps/DEMO/models
```

Additional notes:
* Running any demos in `rmf_demos_gz` will be smooth and have no missing models, as it looks for models in `install/rmf_demos_maps/share/rmf_demos_maps/maps/DEMO/models`

* However during the first run using `rmf_demos_ign` (after clearing up `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models`), maps like `airport_terminal` will require downloading a couple of additional models that have spaces in their names, for example `Oak tree`, `Pine tree`, `Cafe Table`, etc. This is due to the URI used in ignition's world sdf files, where these model names contain `%20` in place of space.
  * This downloading will only take a few seconds, thankfully, and it only happens during the first time
  * Its effects can be seen when checking `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models`, there will be duplicated models now, but with `%20` instead of spaces, for example, `'oak tree'` and `oak%20tree`
  * In the subsequent runs this downloading doesn't happen anymore

I would suggest going forward to merge this PR, while this particular edge case has its own ticket.

luca-della-vedova:
>     * However during the first run using `rmf_demos_ign` (after clearing up `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models`), maps like `airport_terminal` will require downloading a couple of additional models that have spaces in their names, for example `Oak tree`, `Pine tree`, `Cafe Table`, etc. This is due to the URI used in ignition's world sdf files, where these model names contain `%20` in place of space.
>       
>       * This downloading will only take a few seconds, thankfully, and it only happens during the first time
>       * Its effects can be seen when checking `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models`, there will be duplicated models now, but with `%20` instead of spaces, for example, `'oak tree'` and `oak%20tree`
>       * In the subsequent runs this downloading doesn't happen anymore
> 
> 
> I would suggest going forward to merge this PR, while this particular edge case has its own ticket.

Interestingly this seems to be a change of behavior in `ign-fuel-tools` and is not really related to the world files that are the same.
Looking at the debug from Ignition I saw a quite explanatory debug statement [here](https://github.com/ignitionrobotics/ign-fuel-tools/blob/ign-fuel-tools7/src/LocalCache.cc#L604-L611):

```
[ign-24] [Dbg] [LocalCache.cc:604] Model [oak%20tree] loading resource from another model, named [Oak tree]. On Blueprint (ign-fuel-tools 3) and Citadel (ign-fuel-tools 4), [Oak tree] is ignored. From Dome (ign-fuel-tools 5), [oak%20tree] will be used. If [Oak tree] is not a model belonging to owner [openrobotics], fix your SDF file!
```

I guess there is no real fix than this other than removing the space in the SDF file, which would cause quite a lot of repercussions.

Anyway I had a try at the PR and it looks great, my only worry is about what will happen if models are updated in fuel, now users need to worry not only of clearing their `build/` and `install/` directories, but also their `~/.ignition/fuel` directories.
This is only a side effect of the fix but I wonder if it's even possible at all to have pit_crew check models for updates, other than their existence (maybe @methylDragon has more info on the topic?)

methylDragon:
I'm don't think the current arrangement allows for checking of model updates :/

The model metadata caching stops querying metadata once it is cached. I'm also not sure if fuel causes models to bubble to the top of the model list if updated (I think no?) If they do bubble to the top then the model cache building logic will fail because it'll terminate once it finds a model that was previously cached.

If we want to be able to update models, I suppose the only way is to rebuild the cache from scratch, or force a re-download of each model. Either that or, if the author name is findable from the SDF file (and is accurate), implement a function to query fuel directly for the model's metadata and make the download decision there...

luca-della-vedova:
> 

I see, is there a user friendly (i.e. doable through `colcon`) way to force re-download of models? I'm just a bit worried of the case of users having to delete both their build folder and a hidden folder in their home to be guaranteed to get the latest version

luca-della-vedova (approved):

methylDragon:
Mmm, I suppose it might be possible to run another script in the build script to check the last modified time of models in the model path, and redownload if stale?

aaronchongth:
Sorry for missing this discussion, we can move it to another ticket regarding force-redownloading. I'll merge this PR to fix the issue first. Thanks all for the help!

PR #418, opened 2022-02-03
Feature/render OpenStreetMap tiles

codebot:
This is a mega-PR that allows Traffic Editor to work in a global coordinate system: WGS84 latitude/longitude. To make sense of where you are in the world the editor will stream OpenStreetMap tiles into the UI when you're editing a WGS84-based map.

OpenStreetMap tiles are cached locally in a subdirectory in `~/.cache`. The size of this cache is periodically calculated. At some point in the future we could consider auto-deleting old tiles, but at the moment in our current use case it seems to stay around 100-150 MB given "typical" amounts of scrolling/panning/zooming, which doesn't seem so bad.

To choose between this new "geo-referenced" mode and the "classic" Traffic-Editor coordinates of "pixels on a canonical building drawing", there is a dialog box that pops up when you create a new building map.


cnboonhan (approved):
looks great to me!

PR #423, opened 2022-03-07
fix cmake ament_index_cpp dependency for rolling

codebot:
Haven't tested this yet on Jammy, but I think this might fix the `rolling` builds which are currently failing on Ubuntu Jammy. At least, this doesn't break `galactic` on Ubuntu Focal.


Yadunund (approved):
LGTM 😄

PR #427, opened 2022-03-14
Update changelog

codebot:
Prepare for 1.5.0 release


rmf_traffic_editor_test_maps/CHANGELOG.rst
```diff
@@ -2,6 +2,9 @@
 Changelog for package rmf_traffic_editor_test_maps
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+1.5.0 (2022-03-13)
+------------------
+
```
marcoag, line 7: In previous RMF packages, and as an agreement with @mxgrey, we wouldn't add the entry to the `CHANGELOG.md` if there's nothing new added to the package. However since it's actually getting released but with nothing new I think it makes sense to add it and leave it empty like this.

marcoag (approved):

PR #432, opened 2022-06-16
[Humble] Move to collections.abc for crowdsim

luca-della-vedova:
### Fixed bug

Since Python 3.10 (default in 22.04) `collections` is deprecated hence rmf_traffic_editor fails to build maps.

### Fix applied

Move to `collections.abc`

### DIFF
--- a/rmf_building_map_tools/building_crowdsim/navmesh/file_writer.py
+++ b/rmf_building_map_tools/building_crowdsim/navmesh/file_writer.py
@@ -1,4 +1,4 @@
-from collections import Iterable
+from collections.abc import Iterable
 import sys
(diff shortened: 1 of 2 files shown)


aaronchongth (approved):

Yadunund:
To keep backwards compatibility, should we check version and then call the correct import?
```
>>> sys.version_info
sys.version_info(major=3, minor=8, micro=10, releaselevel='final', serial=0)
```

aaronchongth:
It works on Python `3.8.10` for Ubuntu `20.04`, and a deprecation warning is also printed when using the old `import` call, I think this is fine.

PR #436, opened 2022-06-23
Added a dispensable field for models

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

### Implemented feature

Models have a new parameter field `dispensable`, which will have `_dispensable` appended to their model names when the world file is generated.

### DIFF
--- a/rmf_traffic_editor/gui/model.cpp
+++ b/rmf_traffic_editor/gui/model.cpp
@@ -87,6 +87,18 @@ void Model::from_yaml(
     is_static = data["static"].as<bool>();
   else
     is_static = true;
+
+  if (data["dispensable"])
+  {
+    is_dispensable = data["dispensable"].as<bool>();
+  }
+  else
+  {
+    qWarning(
+      "parsed a deprecated .building.yaml, models should have dispensable "
+      "defined.");
+    is_dispensable = false;
+  }
 }
 
 YAML::Node Model::to_yaml(const CoordinateSystem& coordinate_system) const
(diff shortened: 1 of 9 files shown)


rmf_traffic_editor/gui/model.cpp
```diff
@@ -87,6 +87,18 @@ void Model::from_yaml(
...
+
+  if (data["dispensable"])
+  {
+    is_dispensable = data["dispensable"].as<bool>();
+  }
+  else
+  {
+    qWarning(
+      "parsed a deprecated .building.yaml, models should have dispensable "
+      "defined.");
```
luca-della-vedova, line 99: Wondering if this can be silenced or at least only printed once? I saw a common way to only print once in Gazebo that is to include a `static bool` variable in the `else` statement that is set when the warning is printed, although it would have the drawback of persisting through multiple map reloads without application restart.
It seems the new field is mostly needed moving forward so behavior shouldn't be broken if the map is missing the field? Loading a map with a lot of models (i.e. airport_terminal in `rmf_demos`) now prints a lot of warnings that pollute output quite a bit.

aaronchongth: :thinking: At the same time, the default value does not really break any behavior, you are right. There should be no harm in not printing this warning, as long as any future save will save that parameter to be used in the future.

aaronchongth: Removed in https://github.com/open-rmf/rmf_traffic_editor/pull/436/commits/88c9d3c61389b063e36aea017c7d746c46f362c6

luca-della-vedova (dismissed):
Thanks! Looks reasonable to me, we should probably have a followup PR in `rmf_demos` to update the demo maps

PR #442, opened 2022-09-12
building_map_server crashes when level scale is not defined

Yadunund:
In some cases where we create `.building.yaml` files without floorplans images and also without any scale parameters, `building_map_server` will crash when trying to access `level.transform.scale`. This PR aims to address this by only attempting to populate image properties if the file exists.


luca-della-vedova (approved):
Thanks for the fix!

PR #443, opened 2022-09-17
[CI] Created build CI  

orensbruli:
Signed-off-by: Esteban Martinena <orensbruli@gmail.com>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,15 @@
+name: build
+on:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  build_and_test:
+    name: rmf_traffic_editor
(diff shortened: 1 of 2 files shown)


luca-della-vedova:
I noticed test failures due to QT on both this action and, as a consequence, the nightly.

It seems the workflow here uses [a flag](https://github.com/open-rmf/rmf_traffic_editor/blob/main/.github/workflows/ci.yaml#L42) for colcon test:

```
QT_QPA_PLATFORM=offscreen colcon test --packages-select rmf_traffic_editor
```

 I wonder if it would fix the issue?

orensbruli:
> Edit: Noted it's already in the workflow, I wonder if either the flag is not being passed to `colcon test` or there is some other difference in the image I'm not catching

For some reason, it's not setting the QT specific env variable. Investigating.
https://github.com/open-rmf/rmf_traffic_editor/actions/runs/3071883178/jobs/5012288157#step:7:42

luca-della-vedova:
I believe this is superseded by #457.

PR #448, opened 2022-11-08
Updated deprecated GzScene3D to Minimal Scene plugins

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

### Fix applied

I'm not really sure which version of `Gazebo` this project is targeting, but from `Fortress` we deprecated the `GzScene3D` plugin and we should be using these set of plugins.


luca-della-vedova:
Superseded by https://github.com/open-rmf/rmf_traffic_editor/pull/483

PR #463, opened 2023-05-03
Add white wall texture 

Yadunund:
This PR adds a plain white wall texture png. I think it's worth adding this file to the repo as its a common texture at most sites.


codebot (approved):
:+1: :art:

PR #464, opened 2023-05-21
Switch to rst changelogs

Yadunund:
This is to align with [REP 132](https://www.ros.org/reps/rep-0132.html) and will greatly improve the ease of updating changelogs in the future as they will be auto-populated when running [catkin_generate_changelog](https://docs.ros.org/en/rolling/How-To-Guides/Releasing/Subsequent-Releases.html#updating-changelog)

> Note: The `CHANGELOG.rst` was autogenerated from the existing `CHANGELOG.md` using [catkin_md2rst_changelog](https://github.com/Yadunund/catkin_pkg/blob/yadu/md2rst/src/catkin_pkg/cli/md2rst_changelog.py)


codebot (approved):
LGTM

Issue #469, opened 2023-06-20
[pit crew] Models are downloaded each time when running Gazebo classic simulations


Yadunund:
After building latest `main`, running any `rmf_demos_gz_classic` sim will result in gazebo-11 attempting to download every model each time the simulation is run. Sometimes this takes quite a bit of time. 

Haven't narrowed down the cause of the problem but wonder if changes to the downloaded model path made in [this PR](https://github.com/open-rmf/rmf_traffic_editor/pull/454) may have something to do with it. Esp if gazebo-11 does not check for models from `~/.ignition/fuel/fuel.gazebosim.org`.

mxgrey:
This problem has made gz-classic unusable for me since December of 2022.

IIRC the problem may be related to gz-classic and fuel in some way. I think I was able to replicate the problem using assets unrelated to pitcrew.

PR #470, opened 2023-07-04
Fix/gz classic model dl

aaronchongth:
### Fixed bug

Fixes https://github.com/open-rmf/rmf_traffic_editor/issues/469

### Fix applied

Revert the use of http downloads, as it retains the `model://` URI tags, instead of the full Fuel URI, https://github.com/gazebosim/gz-fuel-tools/issues/77#issuecomment-648548920.


mxgrey (approved):
I did a before and after test with a clean build for each, and this PR worked perfectly. Thank you so much for fixing this, gz-classic is usable again 🎉

PR #476, opened 2023-10-06
Handle geometry collections

arjo129:
## Bug fix

I was trying a clean build of RMF Demos and ran into a type error when trying to compile maps. This PR fixes the type error. I'm not sure what triggered it in the first place TBH.


Yadunund (approved):

Yadunund:
@arjo129 can you ensure commits are signed?

PR #478, opened 2023-10-10
Export lifts to nav graphs

mxgrey:
An upcoming version of `rmf_traffic` will be able to take advantage of lift properties, so we need to export that information into the nav graphs.

### DIFF
--- a/rmf_building_map_tools/building_map/building.py
+++ b/rmf_building_map_tools/building_map/building.py
@@ -352,6 +352,7 @@ class Building:
             g = {}
             g['building_name'] = self.name
             g['levels'] = {}
+            g['lifts'] = {}
 
             if self.coordinate_system == CoordinateSystem.web_mercator:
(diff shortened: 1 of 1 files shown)


TanJunKiat:
Is it intentional for the `nav_graphs` that does not utilize the lifts to have the lift properties parsed in the nav_graph file?

mxgrey:
It's not exactly intentional or unintentional. The lift exists in the world, so it makes sense to parse and recognize it as one of the known lifts. In theory it's possible for a user to add more lanes to the graph after parsing it from a file, for example to the generate connections into lifts as part of some automation pipeline. 

I can acknowledge that this is a low likelihood scenario, but generally speaking I don't want to make assumptions about what information a user might want, and it doesn't do any notable harm to parse the lift information just in case.

TanJunKiat:
Alright!

I initially thought the lift information in the `navgraph` generally meant that this fleet has access to these resources.

PR #483, opened 2024-01-19
Harmonic release and ECS refactor

luca-della-vedova:
### Implemented feature

Needed for https://github.com/open-rmf/rmf_simulation/pull/114.

This PR changes the world template to be up to date for Harmonic, as well as change our plugin simulations to use the ECS pattern (single door / lift system, each door / lift has a component registration system).
I'll try to see if we can do a gz / ign rename as well, but can't confirm that will happen yet

Bonus features:

- [x] Rename `ignition` to `gazebo`.
- [x] Remove gazebo classic pipeline.


mxgrey (approved):
The changes look good. I've tested this against 24.04 and rolling, and everything seems to be working.

Now I'll focus on reviewing https://github.com/open-rmf/rmf_simulation/pull/114 and https://github.com/open-rmf/rmf_demos/pull/206, and then we can merge all three simultaneously once CI is sorted out.

luca-della-vedova:
Uncrustify found a few issues with lines above 80 characters (not sure why it didn't before actually), addressed in https://github.com/open-rmf/rmf_traffic_editor/pull/483/commits/bcc23621f681aabb20e218d1555e70f3a9d250f6.

Admin access is needed to remove the `Required` badge from the `rolling,jammy` workflow since it doesn't exist anymore, and move it to `rolling,noble`. After that we should be good to go

mxgrey (approved):

PR #491, opened 2024-04-04
Speed up pagination for pit_crew

methylDragon:
The default number of models that fuel returns is 20. This PR makes the number of models returned per page 5x larger, meaning fewer calls are made, and pagination completes much faster.

Setting the `per_page` to higher than 100 causes fuel to reject the arg and drop you back down to 10 models per page.

This mostly impacts people building from scratch, since the greatest impact is on initial building of the model cache.


methylDragon:
PS: I don't know why DCO is failing even though I have a signoff.

luca-della-vedova:
Looks good! It seems the DCO expected @google.com but got @gmail.com instead, but I think that's OK.

luca-della-vedova (approved):

PR #492, opened 2024-04-04
Workaround for duplicate models for pit_crew

methylDragon:
Although the invariant for the fuel database where a (model_name-author-name) pair is meant to be unique, a duplicate model still slipped through. This causes pit_crew to halt early, causing an incomplete model cache from being built, at worst missing all models on fuel dated earlier than the duplication when building the model cache from scratch.

This causes any fresh builds of RMF demos (and any other building yamls for that matter) using those missing models to fail to find and download them.

# The Fix
This uses a triplet of (model_name-author_name-upload_time) to distinguish models, instead of (model_name-author_name).

# GIANT NOTE
This is not tested against someone running pit_crew's download against a pre-existing model cache. I also haven't refined this yet. I'm just uploading this to present the idea.

This is also a little messy, since we also used the `ModelNames` named-tuple to fetch and classify models from the local cache AND the local gazebo model directory. This is problematic because those don't have upload times, so there's a data mismatch.

### DIFF
--- a/rmf_building_map_tools/pit_crew/pit_crew.py
+++ b/rmf_building_map_tools/pit_crew/pit_crew.py
@@ -63,7 +63,7 @@ __all__ = [
     "build_and_update_cache",
     "init_logging",
     "PitCrewFormatter",
-    "ModelNames",
+    "ModelTuple",
     "download_model_fuel_tools",
     "sync_sdf"
 ]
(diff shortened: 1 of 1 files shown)


koonpeng:
This addresses the same issue as #490 but in a more complete way. The best approach is still fixing fuel's db, if we can do that then both #490 and this can be closed.

luca-della-vedova:
Merged #490 for a hotfix but reverted it when merging main back to this. I'll start testing it now on a prepopulated cache and see how that works

PR #495, opened 2024-05-29
Fix buildfarm build of rmf_traffic_editor packages

luca-della-vedova:
## Bug fix

Fix uncrustify test on buildfarm.

### Fixed bug

Trying to address [buildfarm failure](https://build.ros2.org/job/Jdev__rmf_traffic_editor__ubuntu_noble_amd64/3/consoleFull).

My hypothesis is that `rmf_utils` was not specified as a test dependency, hence the config file would not be found and when running uncrustify it would actually not pass our custom config file (when looking for `ament_uncrustify` in the log, there is no -c [path to file] argument). Hence my proposed fix is to just add it as a test dependency.
The other line is just a cleanup in the CMakeLists since it still refers to a path in galactic


luca-della-vedova:
Upon further investigation, it seems this would make the build from `UNSTABLE` to green, but the real problematic package is `rmf_traffic_editor_assets`

luca-della-vedova:
https://github.com/open-rmf/rmf_traffic_editor/pull/495/commits/df611c4853ea00853445015ae2dc993778d143e4 changed the `rmf_traffic_editor_assets` from a Python to a CMake package, I added the package to the CI and it was red before this PR green with it so it can also act as a regression test for (unlikely) future regressions

Yadunund (approved):

PR #496, opened 2024-05-31
Fix MultiPolygon not iterable

xiyuoh:
Similar to https://github.com/open-rmf/rmf_traffic_editor/pull/476, `shapely` migration requires a minor [update](https://shapely.readthedocs.io/en/stable/migration.html#multi-part-geometries-will-no-longer-be-sequences-length-iterable-indexable) to our code. Without this PR fix the following error will occur when building certain maps that involve MultiPolygon: `TypeError: 'MultiPolygon' object is not iterable`

### DIFF
--- a/rmf_building_map_tools/building_map/floor.py
+++ b/rmf_building_map_tools/building_map/floor.py
@@ -133,7 +133,7 @@ class Floor:
                         plt.plot(poly_x, poly_y, 'b', linewidth=4)
 
             elif geom.geom_type == 'MultiPolygon':
-                for poly in list(geom):
+                for poly in list(geom.geoms):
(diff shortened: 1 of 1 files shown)


luca-della-vedova (approved):
Thanks for the fix! A shame we couldn't get this to reproduce in the test map to get a test for CI but I can confirm this fixes the issue in some of our more complex maps

Yadunund:
Let's follow up with a PR to test this with the test map?

luca-della-vedova:
The problem is that the issue is very rare and actually doesn't happen in any of our demo scenarios so it's unclear how to replicate it. It's probably some edge case with floor triangulation and polygon subtraction that is not immediately clear to me

PR #503, opened 2024-08-12
Fix reference file name being kept when loading new building (#501)

luca-della-vedova:
### Fixed bug

Backport #501 to Humble


xiyuoh (approved):
LGTM

Issue #506, opened 2024-09-06
[Bug]: Lift Name stored as number instead of string in .building.yaml


cardboardcode:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 22.04

### Open-RMF installation type

Binaries

### Open-RMF version or commit hash

 release-humble-231229

### ROS distribution

Humble

### ROS installation type

Binaries

## **Issue Description** :bug: 
When creating a lift and setting `lift name`, naming the lift a number causes runtime error when feeding the generated `.building.yaml` file from RMF Traffic Editor to `building_map_server`.

![image](https://github.com/user-attachments/assets/7a2a3d8b-e9e7-41bb-8615-6743cd42be34)

The error is reported as such:

```bash
[building_map_server-13] Traceback (most recent call last):
[building_map_server-13]   File "/opt/ros/humble/lib/rmf_building_map_tools/building_map_server", line 33, in <module>
[building_map_server-13]     sys.exit(load_entry_point('rmf-building-map-tools==1.6.2', 'console_scripts', 'building_map_server')())
[building_map_server-13]   File "/opt/ros/humble/lib/python3.10/site-packages/building_map_server/building_map_server.py", line 339, in main
[building_map_server-13]     n = BuildingMapServer(map_path)
[building_map_server-13]   File "/opt/ros/humble/lib/python3.10/site-packages/building_map_server/building_map_server.py", line 49, in __init__
[building_map_server-13]     self.load_building_yaml(map_path)
[building_map_server-13]   File "/opt/ros/humble/lib/python3.10/site-packages/building_map_server/building_map_server.py", line 87, in load_building_yaml
[building_map_server-13]     self.create_map_msg(building)
[building_map_server-13]   File "/opt/ros/humble/lib/python3.10/site-packages/building_map_server/building_map_server.py", line 107, in create_map_msg
[building_map_server-13]     self.map_msg.lifts.append(self.lift_msg(lift_data))
[building_map_server-13]   File "/opt/ros/humble/lib/python3.10/site-packages/building_map_server/building_map_server.py", line 286, in lift_msg
[building_map_server-13]     msg.name = lift.name
[...]

PR #507, opened 2024-09-07
Fix for issue #506

cardboardcode:
### Fixed bug :bug: :red_circle: 

This fix resolves issue #506 which surfaces when a lift name only contains numerals. 

Due to how `.yaml` are generated via C++, numeral keys are stored without the necessary quotation, even when explicit stored as a string.

As a result, when loading the generated `.building.yaml` RMF map file into `building_map_server`, it will fail with the following runtime error:

```bash
File "/opt/ros/humble/local/lib/python3.10/dist-packages/rmf_building_map_msgs/msg/_lift.py", line 180, in name
assert \
AssertionError: The 'name' field must be of type 'str'
```

### Fix applied :fist_left: :green_circle: 

This fix explicitly cast the `lift.name` variable in `building_map_server.py` to a string and avoiding the Assertion Error as highlighted above.

### DIFF
--- a/rmf_building_map_tools/building_map_server/building_map_server.py
+++ b/rmf_building_map_tools/building_map_server/building_map_server.py
@@ -283,7 +283,7 @@ class BuildingMapServer(Node):
         msg = Lift()
         # transformation is already done in Lift class
         msg.ref_x, msg.ref_y = lift.x, lift.y
-        msg.name = lift.name
+        msg.name = str(lift.name)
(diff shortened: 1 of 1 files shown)


luca-della-vedova (approved):
Thanks for the fix!

luca-della-vedova:
Can you sign the commit? https://docs.github.com/authentication/managing-commit-signature-verification/about-commit-signature-verification

Issue #462, opened 2023-04-10
Start up error when waypoint's name is pure numbers


mbb1234567890:
## Bug report

Start up rmf_demo error, because of waypoint's name has pure numbers

### Required information:

- Operating system and version:
  - 
  - Ubuntu 20.04
- OpenRMF installation type:
  - 
  - From source
- OpenRMF version or commit hash
  - 
  - c5910c1a Luca Della Vedova <luca@openrobotics.org> on 22-12-6 at 下午2:35
  - committed by GitHub <noreply@github.com>      Unverified GPG signature
- ROS distribution and version:
  - 
  - rolling
- ROS installation type:
  - 
  - binaries
- Package or library, if applicable:
  - 

### Description of the bug

> [fleet_adapter-14] 2023-04-10 17:37:24.539 line526 INFO config_path=/home/bing/rmf_galactic/install/rmf_demos/share/rmf_demos/config/fangshan/harixRobot_config.yaml, nav_graph=/home/bing/rmf_galactic/install/rmf_demos_maps/share/rmf_demos_maps/maps/fangshan/nav_graphs/0.yaml, user_sime_time=False, graph_build_file_path=/home/bing/rmf_galactic/install/rmf_demos_maps/share/rmf_demos_maps/fangshan/fangshan.building.yaml
[building_map_server-3] [INFO] [1681119444.562163742] [building_map_server]: loading map path: /home/bing/rmf_galactic/install/rmf_demos_maps/share/rmf_demos_maps/fangshan/fangshan.building.yaml
[building_map_server-3] [INFO] [1681119444.570400415] [building_map_server]: opening: /home/bing/rmf_galactic/install/rmf_demos_maps/share/rmf_demos_maps/fangshan/fangshan.png
[building_map_server-3] [INFO] [1681119444.571744718] [building_map_server]: read 15945 byte image
[building_map_server-3] Traceback (most recent call last):
[building_map_server-3]   File "/home/bing/rmf_galactic/install/rmf_building_map_tools/lib/rmf_building_map_tools/building_map_server", line 11, in <module>
[building_map_server-3]     load_entry_point('rmf-building-map-tools==1.6.0', 'console_scripts', 'building_map_server')()
[building_map_server-3]   File "/home/bing/rmf_galactic/install/rmf_building_map_tools/lib/python3.8/site-packages/building_map_server/building_map_server.py", line 339, in main
[building_map_server-3]     n = BuildingMapServer(map_path)
[...]

PR #508, opened 2024-09-07
Fix for issue #462

cardboardcode:
### Fixed bug :bug: :red_circle: 

This fix resolves issue #462 which surfaces when a waypoint name only contains numerals. 

Due to how `.yaml` are generated via C++, numeral keys are stored without the necessary quotation, even when explicit stored as a string.

As a result, when loading the generated `.building.yaml` RMF map file into `building_map_server`, it will fail with the following runtime error:

```bash
File "/opt/ros/humble/local/lib/python3.10/dist-packages/rmf_building_map_msgs/msg/_graph_node.py", line 180, in name
assert \
AssertionError: The 'name' field must be of type 'str'
```

### Fix applied :fist_left: :green_circle: 

This fix explicitly cast the `v[2]['name']` variable in `building_map_server.py` to a string and avoiding the Assertion Error as highlighted above.

### DIFF
--- a/rmf_building_map_tools/building_map_server/building_map_server.py
+++ b/rmf_building_map_tools/building_map_server/building_map_server.py
@@ -195,10 +195,7 @@ class BuildingMapServer(Node):
             graph_msg.name = str(i)  # todo: someday, string names...
             print(f"graph {i} has {len(g['vertices'])} vertices")
             for v in g['vertices']:
-                gn = GraphNode()
-                gn.x = v[0]
-                gn.y = v[1]
-                gn.name = v[2]['name']
+                gn = self.create_graph_node(v[0], v[1], v[2]['name'])
 
                 # add specific params to builidng_map_msg
                 for str_param in ["dock_name",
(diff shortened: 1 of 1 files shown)


luca-della-vedova:
This helps but doesn't quite fix the issue, the same logic is present a few lines before the ones you fixed [here](https://github.com/open-rmf/rmf_traffic_editor/blob/f5d138188678d7adfd2e6d6e908359c555ff1021/rmf_building_map_tools/building_map_server/building_map_server.py#L198-L201) and waypoints with a number as their name still crash because of it.
Can you fix it there as well or, even better, refactor into a small function that generates a `GraphNode` to remove the duplicated code altogether?

luca-della-vedova:
Also will most likely have the same issue as https://github.com/open-rmf/rmf_traffic_editor/pull/507 where commits are required to be signed

cardboardcode:
@luca-della-vedova Thank you for reviewing.

Opted to refactor and abstract the duplicate code to a new function, `create_graph_node`.

luca-della-vedova (approved):

PR #509, opened 2024-09-09
Adding delete button to lift table

TanJunKiat:
## New feature implementation

This PR is to add a delete button in the Lift table in the Traffic editor.

It is a minor feature that just convenience the usage of the Traffic editor.

Screenshot from 2024-09-09 08-57-08

### DIFF
--- a/rmf_traffic_editor/gui/building.cpp
+++ b/rmf_traffic_editor/gui/building.cpp
@@ -96,6 +96,8 @@ bool Building::load(const string& _filename)
 
   if (y["reference_level_name"])
     reference_level_name = y["reference_level_name"].as<string>();
+  else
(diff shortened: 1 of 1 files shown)


luca-della-vedova (commented):
Thanks for the PR! I gave it a quick spin and two main comments:

* Can we have the delete button to be just `Delete` without the `...`? The `...` imply that a new dialog is opened but it's not the case for deleting a lift.
* We should probably do some further cleanup when a lift is deleted. Specifically, I noticed that the `lift_cabin` waypoint property is kept which is not ideal:

image

Edit I also forgot that for DCO to pass we need the commits to be signed, can you do that and force push your earlier commit as well?

PR #511, opened 2024-09-18
Introducing lift delete button

TanJunKiat:
Continuation of #509 

### Implementation summary
- Delete button in `lift_table` to remove lift from map and yam
- Delete any lift cabin way point and any edges connected to said way point

### DIFF
--- a/rmf_traffic_editor/gui/level.cpp
+++ b/rmf_traffic_editor/gui/level.cpp
@@ -498,6 +498,50 @@ bool Level::delete_selected()
   return true;
 }
 
+bool Level::delete_lift_vertex(std::string lift_name)
+{
+  std::vector<int> selected_vertex_idxes;
+  for (int i = 0; i < static_cast<int>(vertices.size()); i++)
+  {
+    if (vertices[i].params.count("lift_cabin"))
+    {
+      if (vertices[i].params["lift_cabin"].value_string == lift_name)
+      {
+        selected_vertex_idxes.push_back(i);
+      }
+    }
+  }
+
+  for (const auto& idx : selected_vertex_idxes)
+  {
+    edges.erase(
+      std::remove_if(
+        edges.begin(),
+        edges.end(),
+        [idx](const Edge& edge)
+        {
+          if (edge.start_idx == idx ||
+          edge.end_idx == idx)
+          {
+            return true;
+          }
+          else
+            return false;
+        }),
+      edges.end());
+
+    for (Edge& edge : edges)
+    {
+      if (edge.start_idx > idx)
+        edge.start_idx--;
+      if (edge.end_idx > idx)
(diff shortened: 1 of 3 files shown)


luca-della-vedova (commented):
Interesting, I was thinking something a lot simpler (just iterating over waypoints and removing the `lift_cabin` property if it matches the deleted lift name, leaving everything else intact) but I can see why it could be helpful to automatically remove the lift vertices and edges that lead into the lift.
I'm very slightly concerned about [this break](https://github.com/TanJunKiat/rmf_traffic_editor/blob/7a55fbd1d695059f3ee5bc40a4601cbb2c27c9f3/rmf_traffic_editor/gui/level.cpp#L510), this will mean that as soon as one waypoint with the matching `lift_cabin` property is found it will stop iterating, which "might" behave unexpectedly if there are more waypoints with that property, which will not be cleaned up properly. Now this is usually not the case but other parts of the code dealing with such property don't have the assumption that there will never be more than one (i.e. here https://github.com/TanJunKiat/rmf_traffic_editor/blob/7a55fbd1d695059f3ee5bc40a4601cbb2c27c9f3/rmf_traffic_editor/gui/lift_dialog.cpp#L441-L449, you can see that a flag is set to denote it as found but the loop doesn't early exit).

luca-della-vedova:
The test failures seem mostly related to uncrustify, you should be able to reproduce with a local `colcon test` run

TanJunKiat:
Right now the implementation is a nested `if` , to check for a `lift_cabin` property, then check if the lift name matches.

I am also worried on the situation where the lift cabin name is changed manually via the property panel which results in unsuccessful deletion of the vertex after deleting the lift.

luca-della-vedova:
Yap I know there are two nested if statements to make sure the lift matches, it's more a "what would happen if there was more than one lift with the requested `lift_name`?", the main other code path I linked would still work but this wouldn't because only the first waypoint would be cleaned up

TanJunKiat:
Alright, what we can do is do remove the `break` and remove all the waypoints that fits the criteria.

luca-della-vedova (dismissed):
Thanks for iterating!

luca-della-vedova:
Ah sorry last hurdle, we need commits to be signed https://docs.github.com/authentication/managing-commit-signature-verification/about-commit-signature-verification, can you do that and force-push to this PR?

luca-della-vedova (approved):

TanJunKiat:
Sorry, still trying to fumble around with the signature. Will get to it asap.

luca-della-vedova (dismissed):

luca-della-vedova (approved):

luca-della-vedova:
Thanks for iterating! (Sign offs can be painful the first time :) )

PR #515, opened 2024-11-07
Cherry-pick Shapely 1.8.0 deprecations (#496, #476)

luca-della-vedova:
### Fixed bug

Cherry pick #496 to `humble`, _from what I understand_ the version that is shipped in [Ubuntu 22.04 is 1.8.0](https://packages.ubuntu.com/jammy/python3-shapely). Also from what I understand shapely 1.8.0 prints a deprecation warning https://shapely.readthedocs.io/en/stable/migration.html#multi-part-geometries-will-no-longer-be-sequences-length-iterable-indexable so this should be OK to backport (and avoid issues for people that might have a newer version of shapely but are using the `humble` repos).

Note that I don't have a 22.04 machine so I couldn't really test this

### DIFF
--- a/rmf_building_map_tools/building_map/floor.py
+++ b/rmf_building_map_tools/building_map/floor.py
@@ -133,7 +133,7 @@ class Floor:
                         plt.plot(poly_x, poly_y, 'b', linewidth=4)
 
             elif geom.geom_type == 'MultiPolygon':
-                for poly in list(geom):
+                for poly in list(geom.geoms):
                     self.triangulate_polygon(poly, triangles)
 
             elif geom.geom_type == 'GeometryCollection':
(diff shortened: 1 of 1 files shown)


xiyuoh (approved):
I can verify that the deprecation warning is printed when building the current `humble` branch in a 22.04/Humble distrobox:
```
/home/xiyuoh/ws_osrf/rmf_ws/install/rmf_building_map_tools/lib/python3.10/site-packages/building_map/floor.py:144: ShapelyDeprecationWarning: Iteration over multi-part geometries is deprecated and will be removed in Shapely 2.0. Use the `geoms` property to access the constituent parts of a multi-part geometry.
  for item in geom:
```

From experience our deployment partners using RMF Humble benefited from the fix so I'm good with this cherry pick. Also tested this fix in the same workspace and there were no errors or warnings.

We should backport https://github.com/open-rmf/rmf_traffic_editor/pull/476 as well since they're targeting the same `shapely` update, and it is also the line that the above warning is complaining about.

luca-della-vedova:
Very good point, I missed that https://github.com/open-rmf/rmf_traffic_editor/pull/515/commits/7d67455c7a5bb077612d61883b3a4f300aafd143

PR #518, opened 2024-12-12
Allowing deletion of lift cabin vertices

TanJunKiat:
### Background and motivation

At the current state, when lift cabin vertices are generated, users are unable to directly delete them from the `traffic editor`. The way to go around currently is to manually go into the `building.yaml` file to delete them, but that might lead to undesirable results like failed deletion of a dangling edge or indexing issues.

This PR introduces the ability to delete lift cabin vertices. Users will be able to directly delete the lift cabin vertices from the `traffic editor` itself and have the application automatically re-index all the vertices as well as take care of any potential dangling edges.

### Feature implementation

The implementation is straightforward.

When a deletion command is issued, the algorithm will check if there are any lanes connected to the selected vertex at the viewport level. As usual, an error message will surface if the vertex is used in any edge or polygon.

If the vertex is free, the algorithm will then check if it is a `lift cabin` vertex, and if true, will proceed to purge all the vertices that share the same `lift cabin` parameter with the same `value_string`, in this case, will be the lift's name.

### DIFF
--- a/rmf_traffic_editor/gui/editor.cpp
+++ b/rmf_traffic_editor/gui/editor.cpp
@@ -975,7 +975,8 @@ void Editor::keyPressEvent(QKeyEvent* e)
       {
         // Get the selected vertex index
         int selected_vertex_idx = -1;
-        for (int i = 0; i < static_cast<int>(building.levels[level_idx].vertices.size()); i++)
+        for (int i = 0;
+          i < static_cast<int>(building.levels[level_idx].vertices.size()); i++)
         {
           if (building.levels[level_idx].vertices[i].selected)
           {
@@ -987,7 +988,8 @@ void Editor::keyPressEvent(QKeyEvent* e)
         if (selected_vertex_idx != -1)
         {
           // Get the selected vertex
-          const auto v = building.levels[level_idx].vertices[selected_vertex_idx];
+          const auto v =
+            building.levels[level_idx].vertices[selected_vertex_idx];
           auto it = v.params.find("lift_cabin");
           if ((it != v.params.end()))
           {
--- a/rmf_traffic_editor/gui/editor.cpp
+++ b/rmf_traffic_editor/gui/editor.cpp
@@ -973,6 +973,29 @@ void Editor::keyPressEvent(QKeyEvent* e)
     case Qt::Key_Delete:
       if (building.can_delete_current_selection(level_idx))
       {
+        // Get the selected vertex index
+        int selected_vertex_idx = -1;
+        for (int i = 0; i < static_cast<int>(building.levels[level_idx].vertices.size()); i++)
+        {
+          if (building.levels[level_idx].vertices[i].selected)
+          {
+            selected_vertex_idx = i;
+            break;
+          }
+        }
+
+        if (selected_vertex_idx != -1)
+        {
+          // Get the selected vertex
+          const auto v = building.levels[level_idx].vertices[selected_vertex_idx];
+          auto it = v.params.find("lift_cabin");
+          if ((it != v.params.end()))
+          {
+            // Remove all vertices with a lift cabin property that matches the selected vertex
+            building.purge_lift_cabin_vertices(v.lift_cabin());
+          }
+        }
+
         undo_stack.push(new DeleteCommand(&building, level_idx));
         create_scene();
       }
--- a/rmf_traffic_editor/gui/level.cpp
+++ b/rmf_traffic_editor/gui/level.cpp
@@ -353,15 +353,6 @@ bool Level::can_delete_current_selection()
   if (vertex_used)
     return false;// don't try to delete a vertex used in a shape
 
-  /// check if this is a lift_cabin waypoint
-  const auto v = vertices[selected_vertex_idx];
-  auto it = v.params.find("lift_cabin");
-  if ((it != v.params.end()))
-  {
-    printf("This waypoint is used by a lift cabin!!");
-    return false;
-  }
-
   return true;
 }
 
--- a/rmf_traffic_editor/gui/building.cpp
+++ b/rmf_traffic_editor/gui/building.cpp
@@ -810,6 +810,14 @@ Lift Building::get_lift(const std::string& name) const
   return Lift();
 }
 
+void Building::purge_lift_cabin_vertices(std::string lift_name)
+{
+  for (auto& level : levels)
+  {
+    level.delete_lift_vertex(lift_name);
+  }
+}
+
 void Building::add_constraint(
   const int level_idx,
   const QUuid& a,
--- a/rmf_traffic_editor/gui/building.h
+++ b/rmf_traffic_editor/gui/building.h
@@ -198,6 +198,8 @@ public:
 
   Lift get_lift(const std::string& name) const;
 
+  void purge_lift_cabin_vertices(std::string lift_name);
+
 private:
   std::string filename;
 };


luca-della-vedova (commented):
Thanks for the PR! I tried out the behavior and I have a few questions / feedback.

The main one is about the behavior, the filtering to make sure the vertex is not used in any edge currently only happens on the level that is being rendered, other levels will _not_ be checked and all edges will be cleaned up automatically.
I find this behavior a bit confusing, on one end we refuse to delete the waypoint if it is connected to an edge, and show a warning to the user, on the other as long as the user goes on a level that has no edge we have no problem we recursively deleting all the other edges on other levels.
Is this intentional? My instinct would have been to extend the check for `can_delete_current_selection` to make sure that, if the vertex is a lift vertex, _all_ other lift vertices have no edges connected to them, or in alternative only delete the vertex on the current level (but I'm guessing that will come with its own challenges?)

Then a small nit that I didn't add in line because it happens in an unrelated part of the code, we should change the text in the dialog that spawns when trying to delete a vertex that cannot be deleted to remove mentions of lifts if we will be allowing deleting lift vertices.

TanJunKiat:
> Is this intentional? My instinct would have been to extend the check for `can_delete_current_selection` to make sure that, if the vertex is a lift vertex, _all_ other lift vertices have no edges connected to them, or in alternative only delete the vertex on the current level (but I'm guessing that will come with its own challenges?)

I would say it depends on the intention of the user.

If the user truly intends to perform a deletion of a vertex, informing that the vertex is used and forcing users to delete all the connecting entities before allowing them to delete, the whole process can be a very tedious and roundabout way.

But at the same time, I can vaguely understand the original intent, to warn the users that the vertex is still being used and deleting it might cause irreversible damage.

The current implementation makes it convenient for users if they are working on a large amount of levels, so much that it will take a lot of time to delete all the connecting lanes. We can of course do what as suggeted, meaning to check if other levels are also clear of connections before deleting.

Another option we can go about is to open a dialog when the user wants to delete a vertex that is currently in use (not limiting it to lift cabins), and give them an option to either interrupt the delete action, or to proceed and delete all connecting entities

luca-della-vedova:
I can see where you come from but I would argue that at least the current behavior is consistent, specifically this:

> If the user truly intends to perform a deletion of a vertex, informing that the vertex is used and forcing users to delete all the connecting entities before allowing them to delete, the whole process can be a very tedious and roundabout way.

With this PR in _some cases_ we would still delete vertices that are used while in some cases we wouldn't.
That being said, I agree that it can be a bit tedious so I wonder how tricky it would be to implement your last option:
 
> Another option we can go about is to open a dialog when the user wants to delete a vertex that is currently in use (not limiting it to lift cabins), and give them an option to either interrupt the delete action, or to proceed and delete all connecting entities

So at least we could have a consistent behavior. The dialog could even include a list of entities that currently use the vertex so the user knows exactly what is going to happen, but admittedly I'm not sure how big of a change that would be

TanJunKiat:
I can spend some time to look into it.

The last option seems more sensible, in both ease of use and forewarning the consequences of deleting used entities.

I can spend some time to see how it can be implemented while keeping in mind to see if we can use this for normal vertices as well.

TanJunKiat:
Hi @luca-della-vedova 

I've tweaked the implementation and reverted some of the functions.

But the concept is as what we discussed.

Now when the user is deleting a vertex that is used (edges, polygons, or lift cabins), a warning dialog will appear to confirm the action.

When the user confirms the action, if it is a lift cabin, the purging will be done as per the previous commit.

If it is an edge or polygon, it will trigger a purging function to remove the selected entity and the vertex.

rmf_traffic_editor/gui/level.cpp
```diff
@@ -498,6 +498,48 @@ bool Level::delete_selected()
   return true;
 }
 
+bool Level::delete_used_entities()
```
luca-della-vedova, line 501: Would it be possible to change this function signature so it just receives the selected vertex as a parameter rather than having to iterate through all the vertices in the level to find the selected one?

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+DeleteDialog::DeleteDialog(
+  QWidget* parent)
+: QDialog(parent)
+{
+  setWindowTitle("Warning");
+  QVBoxLayout* warning_message_vbox_layout = new QVBoxLayout;
+  warning_message_vbox_layout->addWidget(new QLabel(
+      "This item might be linked to other entities and/or on other levels.\n "
+      "Are you sure you want to delete this item?"));
```
luca-della-vedova, line 33: It would be nice to make the messaging here a bit more specific so the user really knows what is going to happen. I tried the feature and while it works it ended up deleting _a lot_ of entities when I deleted a vertex, this is especially dangerous since `undo` for vertex deletion doesn't really work and users would probably have to reload their project if they delete the wrong vertex.
Since there are different code paths being executed to check usage (from what I see at least lifts, edges and polygons all have different paths) it should be possible to populate a list of entities (or at the very least type of entities with number) that could be affected, so TLDR best scenario something like:

```
This vertex is being used by 10 entities:
* Level [L2]:
  * 1 graph lane(s) for graph index [0]
  * 3 graph lane(s) for graph index [1]
  * 2 walls
  * 1 lift with lift name [lift_1]
  * 1 polygon
*  Level [L2]
  * [...] 
```

Slightly less informative but still better than current

```
This vertex is being used by 10 entities:
* Level [L1]:
  * 5 edges
  * 1 lift
  * 1 polygon
*  Level [L2]
  * [...] 
```

Least informative

```
This vertex is being used by 10 entities:
```

luca-della-vedova (commented):
Looks a lot better! And I feel it will definitely be a helpful feature. Sometimes deleting a short edge can be quite painful if not outright impossible if all you can select is the vertices at its ends.

rmf_traffic_editor/gui/editor.cpp
```diff
@@ -978,14 +979,41 @@ void Editor::keyPressEvent(QKeyEvent* e)
...
+              building.purge_lift_cabin_vertices(v.lift_cabin());
+            }
+          }
+
+          building.levels[level_idx].delete_used_entities();
+          undo_stack.push(new DeleteCommand(&building, level_idx));
+          create_scene();
+        }
+        else
+          building.clear_selection(level_idx);
```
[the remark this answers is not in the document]
TanJunKiat: This is because the selection is not cleared properly.

This can be fixed by adding these 3 lines:
```
clear_current_tool_buffer();
update_property_editor();
create_scene();
```

This clears the tool selection, updates the property panel (clears it because the tool selection is cleared), and updates the scene to visually unselect the entity.

rmf_traffic_editor/gui/level.cpp
TanJunKiat: Yes, my mistake.

The selected vertex is already calculated before the call of this function, so it is redundant to do another loop to find the selected one.

rmf_traffic_editor/gui/delete_dialog.cpp
TanJunKiat: I have made the neccessary changes to achieve something similar to the best scenario.

But it is tailored to two different scenarios.

For lift cabin, it will inform the user that the vertex is used across different levels (may or may not have lanes connected to them) and show the different levels that the lift has access to.

For normal edges and polygons, it is not possible to have inter-level references. So when a user wants to delete a used vertex, the warning will show all the different types of edges and polygons that are currently using said vertex.

TanJunKiat: Screenshot 2024-12-18 at 10 29 00 AM
Screenshot 2024-12-18 at 10 32 16 AM

rmf_traffic_editor/gui/editor.cpp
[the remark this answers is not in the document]
luca-della-vedova: Awesome!

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,126 @@
...
+  if ((it != v.params.end()))
+  {
+    auto lift = building.get_lift(v.lift_cabin());
+    for (const auto& level : building.levels)
+    {
+      if (level.elevation <= lift.highest_elevation &&
+        level.elevation >= lift.lowest_elevation)
+      {
+        lift_levels.push_back(level.name);
+      }
```
luca-della-vedova, line 51: I'm not quite sure this is the right way to check for the lift. Specifically, lifts can have levels that are skipped in-between, in which case this prints the wrong message.
To reproduce, try the following in the hotel world;

* Delete a lift vertex from one of the lifts
* Disable L2 visit in the lift as below:
image

* Add the lift waypoints
* Notice how L2 does _not_ get a lift waypoint
* Delete a waypoint from L3, note how it will say that there is a waypoint in L2 while that is not the case

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,126 @@
...
+        QString("This item is a lift cabin and is used in %1.\n "
+        "Are you sure you want to delete this item?")
+        .arg(QString::fromStdString(s))));
+  }
+  else
+  {
+    std::string s;
+
+    auto num_used_edges = level.edges_with_vertex(selected_vertex_idx).size();
+    std::unordered_map<std::string, int> edge_types;
```
luca-della-vedova, line 68: Nit, when printing to the user it's nice to have a deterministic order, I would suggest using `std::map` here so they are always in alphabetical order, rather than a non deterministic hash map based one.

rmf_traffic_editor/gui/editor.cpp
```diff
@@ -978,14 +979,49 @@ void Editor::keyPressEvent(QKeyEvent* e)
...
+        for (int i = 0;
+          i < static_cast<int>(building.levels[level_idx].vertices.size());
+          i++)
+        {
+          if (building.levels[level_idx].vertices[i].selected)
+          {
+            selected_vertex_idx = i;
+            break;
+          }
+        }
```
luca-della-vedova, line 993: Note to self but let's leave it to a followup, it is actually possible to do Ctrl-Click to select multiple vertices but the codebase only finds the first occurrence, we should consider whether to disallow selecting multiple vertices or make sure deletion can operate on an array of selected elements.

rmf_traffic_editor/gui/editor.cpp
```diff
@@ -978,14 +979,49 @@ void Editor::keyPressEvent(QKeyEvent* e)
...
+            auto it = v.params.find("lift_cabin");
+            if ((it != v.params.end()))
+            {
+              // Remove all vertices with a lift cabin property that matches the selected vertex
+              building.purge_lift_cabin_vertices(v.lift_cabin());
+            }
+
+            building.levels[level_idx].delete_used_entities(
+              selected_vertex_idx);
+            undo_stack.push(new DeleteCommand(&building, level_idx));
```
luca-della-vedova, line 1012: Note to self, this is broken and we should either fix it or disable it

rmf_traffic_editor/gui/level.cpp
```diff
@@ -542,6 +571,28 @@ bool Level::delete_lift_vertex(std::string lift_name)
   return true;
 }
 
+std::vector<Edge> Level::edges_with_vertex(int vertex_idx) const
+{
+  std::vector<Edge> result;
+  for (const auto& edge : edges)
+  {
+    if (edge.start_idx == vertex_idx || edge.end_idx == vertex_idx)
```
luca-della-vedova, line 579: I thought you added a `contains` function that can be used here instead?

luca-della-vedova (commented):
Few small nits (and a few notes to self on things we should address in the future but are out of the scope of this PR) then looks good to me!

rmf_traffic_editor/gui/level.cpp
TanJunKiat: Amended.

rmf_traffic_editor/gui/editor.cpp
TanJunKiat: I think the selection of multiple vertices is required to collinear the selected vertices, so disabling it might be tough.

rmf_traffic_editor/gui/delete_dialog.cpp
TanJunKiat: I made changes to use the `LevelDoorMap` instead.

But that also made me realised that when the door of that level is removed, the lift vertex at that level still persist.

Might want to include a cleanup action to remove that vertex.

rmf_traffic_editor/gui/delete_dialog.cpp
TanJunKiat: Changed to std::map instead as suggested.

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,124 @@
...
+    warning_message_vbox_layout->addWidget(new QLabel(
+        QString("This item is a lift cabin and is used in %1.\n "
+        "Are you sure you want to delete this item?")
+        .arg(QString::fromStdString(s))));
+  }
+  else
+  {
+    std::string s;
+
+    auto num_used_edges = level.edges_with_vertex(selected_vertex_idx).size();
```
luca-della-vedova, line 65:
```suggestion
```

This seems unused and iterates over all the edges in the level so it's quite expensive

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,124 @@
...
+      edge_types[edge.type_to_string()]++;
+    }
+    for (const auto& edge_type : edge_types)
+    {
+      s = s + " - " + std::to_string(edge_type.second) + " " + edge_type.first +
+        " edge(s) \n";
+    }
+
+    auto num_used_polygons =
+      level.polygons_with_vertex(selected_vertex_idx).size();
```
luca-della-vedova, line 78:
```suggestion
```
This also seems unused

rmf_traffic_editor/gui/delete_dialog.h
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2019-2021 Open Source Robotics Foundation
```
luca-della-vedova, line 2: Nit, can you change the copyrights on the new files to 2024?

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "delete_dialog.h"
+#include <algorithm>
+#include <QtWidgets>
+using std::vector;
+using std::string;
+
+
+DeleteDialog::DeleteDialog(
+  QWidget* parent,
+  Building& building,
```
luca-della-vedova, line 27: Can this be changed to `const Building&`? This dialog shouldn't change the building

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,124 @@
...
+  if ((it != v.params.end()))
+  {
+    auto lift = building.get_lift(v.lift_cabin());
+
+    for (const auto& level_door : lift.level_doors)
+    {
+      lift_levels.push_back(level_door.first);
+    }
+
+    std::string s = lift_levels[0];
```
luca-della-vedova, line 51: Spooky, this segfaults if `lift_levels` is empty, which can happen if a user disables all floors and then tries to delete a vertex

rmf_traffic_editor/gui/delete_dialog.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+
+
+  setWindowTitle("Warning");
+  QVBoxLayout* warning_message_vbox_layout = new QVBoxLayout;
+
+  // check if vertex is a lift cabin
+  std::vector<std::string> lift_levels;
+  auto level = building.levels[level_idx];
+  const auto v = level.vertices[selected_vertex_idx];
```
luca-della-vedova, line 40:
```suggestion
  const auto& level = building.levels[level_idx];
  const auto& v = level.vertices[selected_vertex_idx];
```

This would create a whole copy of a level / vertex behind the scenes when spawning the dialog, as well as not be a compile error if the level is edited

rmf_traffic_editor/gui/delete_dialog.h
```diff
@@ -0,0 +1,51 @@
...
+#include <QDialog>
+#include <QObject>
+#include <vector>
+#include <string>
+#include <utility>
+#include <map>
+#include "building.h"
+class QLineEdit;
+class QListWidget;
+class QLabel;
```
luca-della-vedova, line 30: There is a lot of unnecessary headers here (this should only contain what's needed in the header file, what's needed in the cpp can just go in there)

Also do we need to actually declare these QT classes as you are doing?

luca-della-vedova (commented):
Last (hopefully) round, more into the details of the implementation / code and a corner case that causes a segfault in the current implementation

rmf_traffic_editor/gui/delete_dialog.h
TanJunKiat: Removed unused headers.

rmf_traffic_editor/gui/delete_dialog.cpp
TanJunKiat: I've changed the string concatenating logic. It should be safe now.

rmf_traffic_editor/gui/delete_dialog.cpp
TanJunKiat: Removed redundant lines.

rmf_traffic_editor/gui/delete_dialog.cpp
TanJunKiat: Removed redundant lines.

luca-della-vedova:
> Another side note, is it okay to make the `lift cabin` property uneditable?
> 
> If the value is to be edited manually and incorrectly, the behaviour might be undesirable.

No strong feelings about this, there seem to be a lot of other improvements that could be made there, we can leave them for a followup (for example I just noticed that renaming a lift will not rename its cabin waypoints which might also cause unintended side effects, the lift of lifts is not updated, etc)

luca-della-vedova (approved):

TanJunKiat:
> > Another side note, is it okay to make the `lift cabin` property uneditable?
> > If the value is to be edited manually and incorrectly, the behaviour might be undesirable.
> 
> No strong feelings about this, there seem to be a lot of other improvements that could be made there, we can leave them for a followup (for example I just noticed that renaming a lift will not rename its cabin waypoints which might also cause unintended side effects, the lift of lifts is not updated, etc)

Yea, I will try to fix these minor bugs when I have the time.

Thanks for the review on this PR.

PR #520, opened 2024-12-19
Add a conditional check for level transformation

TanJunKiat:
### Fixed bug

From PR https://github.com/open-rmf/rmf_traffic_editor/pull/513, the `compute_transform` function is modified to take into consideration level rotations when doing inter-level transformation.

But a segmentation fault will occur when a new level is created because there are no fiducials.

### Fix applied

I've added a conditional check to make sure that the fiducials are 2 or more, before allowing the function to continue.


luca-della-vedova (approved):
Ouch yes that `[0]` :skull:, thanks!

PR #521, opened 2024-12-27
Support robot types with organization namespaces

luca-della-vedova:
## New feature implementation

Goes together with https://github.com/open-rmf/rmf_demos/pull/272

### Implemented feature

This PR adds support for robots that have an organization namespace in their name and removes the namespace if so, following the same logic as normal models.
~~It also moved the dispenser / ingestors from `OpenRobotics` to `Open-RMF` organization since that is where they are uploaded.~~ Moving has been reverted


mxgrey (approved):
I tweaked this just a little so that it only filters out namespaces of `Open-RMF` (not case sensitive) just in case any user is already using org names.

PR #522, opened 2025-01-02
Fix pit crew missing models and authorship logic

luca-della-vedova:
### Fixed bug

Ref https://github.com/open-rmf/rmf/discussions/578#discussioncomment-11713951, fuel pipeline wasn't used for some time, but while fixing it I bumped into several other issues that I bundled in this PR.

### Fix applied

A series of issues in pit_crew:

* The root cause of the issue, if a model with a different organization name is uploaded `pit_crew` will report it as downloadable, the pipeline will later try to download it but fail. Fixed the `downloadable` generation logic https://github.com/open-rmf/rmf_traffic_editor/pull/522/commits/d392b76d363d78933c8ea25daefdb4a3bf52b9a8.
* The above didn't quite bubble up because in the bare http pipeline (which is what we use) we check a _tuple_ for true / false, so returning `(False, None)` actually evaluates to True, fixed in https://github.com/open-rmf/rmf_traffic_editor/commit/8c3f40c4d5c6b69b1f5578b14f119ac616d4c76b. 
* Now of course the fuel pipeline uses `ign` instead of `gz` which doesn't exist anymore, https://github.com/open-rmf/rmf_traffic_editor/commit/1e6c04e23bcefc5eaec1bfcc1260cd1fa7bbf457 for the command and https://github.com/open-rmf/rmf_traffic_editor/commit/14ee9536fac8f40bfc9f7539170e7b7746ee90bf to fix the cache folder.
* The rest are minor fixes:
  * https://github.com/open-rmf/rmf_traffic_editor/commit/f4d35f9d301bb9fd75c48de156bfff280dfef9d0 the printing was not substituting the variable so would always print "%d".
  * https://github.com/open-rmf/rmf_traffic_editor/commit/0c74e73708405d11ce7e185720838f362f41345a We were enumerating but not using the index, remove the enumerate.
  * https://github.com/open-rmf/rmf_traffic_editor/commit/aaa7a835f0c7da6e52285cb056de0e862e1e4447 the pipeline was throwing an exception when using `max()` on an empty list, change it to a more informative exception.

All off this surfaced because of the following sequence of events: 
* Uploading the [TeleportIngestor](https://app.gazebosim.org/Open-RMF/models/TeleportIngestor) and [TelepotDispenser](https://app.gazebosim.org/Open-RMF/fuel/models/TeleportDispenser) to the `Open-RMF` organization.
* Fuel now has an `Open-RMF/TeleportIngestor` plugin, maps still require `OpenRobotics/TeleportIngestor`.
* The faulty logic makes `OpenRobotics/TeleportIngestor` map to `Open-RMF/TeleportIngestor`.
* Model download with the `http` pipeline fails, but the faulty boolean condition check makes the pipeline progress as usual
* The `fuel` pipeline has the correct check and aborts early.

All of these will be fixed by the above.
Note however, that `TeleportIngestor` and `TeleportDispenser` might be downloaded by fuel in the future, but I'll leave a detailed comment for this in the rmf_demos PR that introduces this risk https://github.com/open-rmf/rmf_demos/pull/272

### DIFF
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -97,7 +97,7 @@ def download_models(
     """Download models for a given input building yaml."""
     # Construct model set
 
-    IGN_FUEL_MODEL_PATH = "~/.ignition/fuel/"
+    IGN_FUEL_MODEL_PATH = "~/.gz/fuel/"
 
     model_set = set()
     stringent_dict = {}  # Dict to tighten download scope
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -180,7 +180,7 @@ def download_models(
             if model_downloaded:
                 break
             else:
-                logger.warning("Retrying %d of 5 times...", i)
+                logger.warning("Retrying %d of 5 times..." % i)
 
         if not model_downloaded:
             logger.error(
--- a/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
+++ b/rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
@@ -176,7 +176,7 @@ def download_models(
                     sync_names=True, export_path=export_path)
             else:
                 model_downloaded = pit_crew.download_model(
-                    model_name, author_name, sync_names=True)
+                    model_name, author_name, sync_names=True)[0]
             if model_downloaded:
                 break
             else:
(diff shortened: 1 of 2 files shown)


mxgrey (approved):
Thanks for the detailed explanation of all the problems and their fixes, that really helped give context for all the surrounding Python code.

mxgrey (approved):
I wonder if there's a way to test the changes without a risk of breaking something for users out there.

What if we upload some empty model to both the openrobotics fuel org and the open-rmf fuel org with the same name and build a test map that contains the model?

ana-GT:
Just a comment: I was having trouble downloading models to run the demonstrations in a Jazzy setup (models were not downloading due to ign not being found, as expected). Using this PR did the trick.

mxgrey:
Thanks for bumping this @ana-GT we should go ahead and merge this so it's in for the kilted release.

PR #528, opened 2025-04-21
Fix to Allow Numerically-Labelled Map

cardboardcode:
### Fixed Bug :bug: 

The bug in this proposed fix is for allowing numerically-labelled map names.

Eg. If map name in `.building.yaml` is `L1`, no error when starting up RMF Traffic. Map can be displayed in `RViz` window.
Eg. If map anme in `.building.yaml` is `1`, RMF Traffic crashes. Map won't be displayed in `RViz` window.

### **Fix Applied** :hammer: 

Explicit-cast level name to string when generating `Level` msg.
Explicit-cast level name associated to lift when generating `Lift` msg.

### DIFF
--- a/rmf_building_map_tools/building_map_server/building_map_server.py
+++ b/rmf_building_map_tools/building_map_server/building_map_server.py
@@ -278,7 +278,7 @@ class BuildingMapServer(Node):
         # transformation is already done in Lift class
         msg.ref_x, msg.ref_y = lift.x, lift.y
         msg.name = str(lift.name)
-        msg.levels = str(level_name) for level_name in lift.level_names
+        msg.levels = [str(level_name) for level_name in lift.level_names]
 
         msg.ref_yaw = lift.yaw
         msg.width = lift.width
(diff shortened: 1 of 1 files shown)


rmf_building_map_tools/building_map_server/building_map_server.py
```diff
@@ -278,7 +278,7 @@ def lift_msg(self, lift):
         # transformation is already done in Lift class
         msg.ref_x, msg.ref_y = lift.x, lift.y
         msg.name = str(lift.name)
-        msg.levels = lift.level_names
+        msg.levels = str(lift.level_names)
```
mxgrey, line 281: This doesn't seem correct to me... I expect `lift.level_names` to be a list, so we would want to convert each element to a `str`, not the entire list into one `str`, e.g.:

```suggestion
        msg.levels = str(level_name) for level_name in lift.level_names
```

I'd appreciate if you can test this out on a multi-level building and let us know how it turns out.

cardboardcode: Noted. Thank you for the suggestion.

Will make the necessary changes and test. 

Will update after I have tested.

cardboardcode: Hi @mxgrey 

I have made the changes. Tested as well by duplicating the first map and naming it `2`. 

Able to view both maps `1` and `2` in RViz.

mxgrey (approved):

PR #535, opened 2025-06-03
Download models through Gazebo

### LINKED ISSUE #534: Use gazebo instead of pit_crew to download models
`pit_crew` was written at a time when we needed a pipeline to download both Gazebo classic and Gazebo models, as well as download models at build time so the installation folder could be redistributed. Now that those requirements are not there anymore, we can change the simulations to just download models at runtime using the better supported method of gazebo fuel URIs.
We could change `pit_crew` to do the following:

* Keep the same input (i.e. models in the form of `OrganizationName/ModelName`).
* At map build time, check if there is such a model on fuel.
* If there is, change the model URI exported to the `.world` file to point to the Fuel model.
* If there isn't, leave it unchanged for Gazebo to find through other methods (i.e. local paths defined in the `GZ_SIM_RESOURCE_PATH` environment variable).
* Stop downloading models.

This should dramatically cut build time for `rmf_demos_maps` since it won't need to download all the models for all the maps, as well as its size making it possible to distribute it as a binary.

luca-della-vedova:
### Implemented feature

Closes #534.

This PR simplifies our world building pipeline by removing model downloading and, instead, substituting fuel server paths for models that are uploaded to fuel, and letting Gazebo itself download them.

Most of `pit_crew` has been removed, the only part left is keeping the cache updated. I considered removing the cache altogether but that would have slowed down builds somewhat.
Previously, for each map we would run the model downloader that would check if the cache needs to be updated, if any models needs to be downloaded, then download them to the user folder.
This was necessary in a time where we needed to support both Gazebo and Gazebo Classic, which had different support for fuel URIs.
Right now however, this functionality is largely unnecessary and model downloading can be left to Gazebo itself.

With this PR (and the matching PR in `rmf_demos`), model building will now look like this:

* A first step in building of a demo package will run one instance of the `building_map_model_downloader` (although it is technically a `model_cache_updater`, happy to rename it). This will only perform the cache maintenance operation that `pit_crew` used to do.
* After that step, all the maps will be built and will use the updated cache file.
* If a model is found on fuel, they will remap the URI from (example) `OpenRobotics/Drawer` to `https://fuel.gazebosim.org/1.0/OpenRobotics/models/Drawer`, otherwise it will leave it as is.
* That's it!

This approach has the following benefits:

* We don't need to download all the models for all the maps at build time anymore, saving significant resources.
* We only need to run one instance of pit_crew for all the maps and not pit_crew only needs to do a single "update cache" operation, making build time scale by a constant factor rather than linearly with the number of maps / models.
* We don't need to have all models available at runtime and can let Gazebo download them when the simulation is ran instead, unblocking distribution of full rmf_demos binaries.

There is only two minor drawbacks to my knowledge:

* Simulations will need internet access for their first run, to allow Gazebo to download the models from fuel.
* The first run of the simulation on a user machine might take longer since Gazebo will need to download the models, the same amount of time is saved on build time but users might be confused by a slower first simulation start time.


rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
```diff
@@ -1,35 +1,46 @@
...
 import yaml
 
 
 __all__ = [
-    "download_models"
+    "update_cache"
 ]
 
+
+class PitCrewFormatter(logging.Formatter):
```
mxgrey, line 20: Are we keeping this name as an homage to `pit_crew`?

rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
```diff
@@ -39,187 +50,137 @@ def __call__(self, parser, namespace, values, option_string=None):
...
-                        Redundant if using ignition fuel tools flag")
 parser.add_argument("-c", "--cache", type=str,
                     default="~/.pit_crew/model_cache.json",
                     help="Path to pit_crew model cache")
-parser.add_argument("-f", "--fuel-tools", action='store_true',
-                    help="Use ignition fuel tools to download models instead "
-                         "of http")
-parser.add_argument("-i", "--include", type=str, default=None,
-                    help="Search this directory first for models.")
-parser.add_argument("-e", "--export-path", type=str, default=None,
```
mxgrey, line 57: Removing these arguments kills backwards compatibility with existing map packages, resulting in build errors if the map packages are not updated.

From what I can tell the changes in this PR make it so that map packages no longer need to call `building_map_model_downloader` at all. What if instead of removing these arguments, we keep them and print out a deprecation warning to `stderr` telling users that they can safely remove the call to `building_map_model_downloader` from their `CMakeLists.txt`?

mxgrey: I've opened #540 with a suggestion for fixing this. I've confirmed that it allows `rmf_demos_maps` to build without needing https://github.com/open-rmf/rmf_demos/pull/307 and prints a deprecation warning.

luca-della-vedova: I'll have a look at what happens with the recommended change WRT cache building. Note that I would still recommend to run the `building_map_model_downloader` and the rmf_demos PR reflects [the recommended change](https://github.com/open-rmf/rmf_demos/pull/307/files#diff-4e7f74c52f6412afcc3703d3ab3740e8b55380687a5d43b602527b92f75e626bR16-R21).
The reason is that while we don't need to download models, we need to query Fuel to know which models are available so we can change them to point to a Fuel asset.
Doing so requires querying Fuel for its entire model database that can take a significant amount of time, hence the change to this script to just build the fuel database cache that can be reused to build consecutive maps.
If we didn't build a cache, we would need every map building invocation to query the fuel database.

mxgrey: Oh the way the diff was laid out made it look to me like we were removing its use entirely.

If the recommendation is to just run `building_map_model_downloader` with no arguments, then I can change the deprecation warning to indicate that, and I'll allow the script to run as normal.

mxgrey: [ac01af6](https://github.com/open-rmf/rmf_traffic_editor/pull/540/commits/ac01af652c12580fd637045a9a1d644c84b0c0c0) changed the warning to reflect what the user should really do, and now it runs the script even when the deprecated arguments are passed in.

luca-della-vedova: We could also just move the (now simplified) logic into the actual package that builds the map and fully deprecate this executable.
I didn't do it because that approach would try to update the cache once for each map when it is really only needed to do it once per package, so we can either have a better build time (this approach) or a more clear deprecation path that fully deprecates the executable (what your original change was).
As an option that could have both benefits perhaps there is a CMake incantation that we could use so we have a `foreach` where in the _first_ iteration we pass an additional flag to tell the map building package to also update its cache and in the other iterations we don't? But I'm not 100% confident in my CMake to be able to do this

mxgrey: > perhaps there is a CMake incantation that we could use

I think the current approach of running `building_map_model_downloader` outside of the loop is the most sensible.

The goal of the PR I opened is to make sure things work okay (even if it's not optimal) for any map packages outside of our control that have copied the `CMakeLists.txt`, and also provide guidance to users on how to make their setup optimal. Personally I don't think we should worry about doing more than that since we have a goal of transitioning to the site editor and releasing a [cmake module package](https://github.com/open-rmf/rmf_site/issues/307) for it. Hopefully users will be able to transition to that before long.

rmf_building_map_tools/building_map_model_downloader/building_map_model_downloader.py
luca-della-vedova: I just brought the code in with as little changes as possible to reduce chance of mistakes / make the diff easier to review, but it's a nice side effect :shrug:, neutral on renaming / keeping it

mxgrey (approved):
I think this pipeline makes a lot of sense. Pit crew served a good purpose when we had to juggle Gazebo Classic and Ignition at the same time, but it's nice to finally shed that need.

One thing that's noteworthy is for a fresh build+install of Open-RMF, the build time will be faster but the first time launching a demo world in Gazebo will take a while as it downloads the models for the first time. Subsequent launches are fine, but first-time users may get confused about why the simulation doesn't appear to be loading.

I think this trade-off is worth it, but we can expect an uptick in issue reports about the simulation not loading, and we should be prepared to check whether it was happening on their first launch, make sure they have an internet connection while launching, and ask them to give it time to load.

PR #538, opened 2025-07-08
Fix bug in angle difference calculation near ±π

kjchee:
### Fixed bug

This PR fixes a bug in the rotation transformation logic where angle differences were calculated using simple subtraction without accounting for angle wraparound in the [-π, π] range.

In the original implementation, the angular difference between two orientations (calculated using atan2) was computed like this:
`double angle_diff = angle1 - angle2;`

This approach fails when the two angles are numerically far apart but spatially very close due to wraparound.
For example:
```
angle1 = 0.9π
angle2 = -0.9π
```
A simple subtraction results in:
`angle_diff = 0.9π - (-0.9π) = 1.8π`

However, geometrically, these two angles are very close — the actual shortest angular distance should be -0.2π .

This leads to large, incorrect relative_rotation_sum, and causes wrong rotation transformation `const double rotation = relative_rotation_sum / rotations.size();`.

### Fix applied

The fix introduces normalization of the angle difference into the range [-π, π]:
```
double diff = rotations[i].first - rotations[i].second;
while (diff <= -M_PI) diff += 2.0 * M_PI;
while (diff >  M_PI)  diff -= 2.0 * M_PI;
relative_rotation_sum += diff;
```

This ensures the smallest angular distance is always used (in the range [-π, π]), resolving errors caused by direct subtraction.

### DIFF
--- a/rmf_traffic_editor/CMakeLists.txt
+++ b/rmf_traffic_editor/CMakeLists.txt
@@ -12,6 +12,7 @@ find_package(Ceres REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(Qt5 COMPONENTS Widgets Concurrent Test Network REQUIRED)
 find_package(yaml-cpp REQUIRED)
+find_package(rmf_utils REQUIRED)
 
 set(CMAKE_BUILD_TYPE RelWithDebInfo)
 # set(CMAKE_VERBOSE_MAKEFILE TRUE)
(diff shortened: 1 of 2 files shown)


mxgrey (approved):
Thanks for catching this!

I've tweaked your PR to use the [`wrap_to_pi`](https://github.com/open-rmf/rmf_utils/blob/5c9d1ae99ec92ef8ac0fec4a15384a6fe3037062/rmf_utils/include/rmf_utils/math.hpp#L26-L35) method from `rmf_utils`. It's effectively the same as what you submitted, but fewer lines of code and therefore fewer opportunities to accidentally introduce mistakes (now or in the future).

Once the CI passes we should be good to merge this!

PR #540, opened 2025-07-15
Restore and deprecate arguments that are no longer needed

mxgrey:
This PR addresses [this feedback](https://github.com/open-rmf/rmf_traffic_editor/pull/535#discussion_r2206599708) that I left on #535


mxgrey:
Note that I change the `"-f", "--force"` to just `"--force"` so it doesn't get mixed up with the deprecated `"-f", "--fuel-tools"` argument.

luca-della-vedova (approved):
Very neat and helpful deprecation path!

PR #544, opened 2026-04-17
Zone editing UI and YAML serialization for GoToZone feature

chart-singapore:
### Implemented feature
This is part of an 8-repo Simple GoToZone feature. Tracked in Stage A of [open-rmf/rmf#726](https://github.com/open-rmf/rmf/issues/726).

Adds zone support to Traffic Editor: a Qt UI for creating, editing, and viewing zones, plus YAML serialization that emits zone structure into building and nav-graph files. Building-map tools now generate zones into the nav-graph yaml.

##### Traffic Editor UI
Added a new zone dialog: 
Screenshot from 2026-04-17 22-08-48

##### Input validation in the zone dialog

- Zone names must be unique within the building.
- Internal-vertex names must be unique within the zone.
- Every internal vertex must have at least one entry lane and one exit lane.
- At least one entry lane and one exit lane must connect to different external waypoints. This prevents "single-door" zones where every entry and exit would go through the same external waypoint and causing deadlock.

### GenAI Use
We follow [OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

- [x] I did not use GenAI

### DIFF
--- a/rmf_building_map_tools/building_map/building.py
+++ b/rmf_building_map_tools/building_map/building.py
@@ -7,6 +7,7 @@ import os
 import requests
 import sqlite3
 import tempfile
+import copy 
 import yaml
 
 from ament_index_python.packages import get_package_share_directory
@@ -20,6 +21,7 @@ from .etree_utils import indent_etree
 from .geopackage import GeoPackage
 from .level import Level
 from .lift import Lift
+from .zone import Zone
 from .material_utils import copy_texture
 from .param_value import ParamValue
 from .passthrough_transform import PassthroughTransform
@@ -162,6 +164,14 @@ class Building:
                          self.coordinate_system)
 
         self.set_lift_vert_lists()
+
+        self.zones = {}
+        if 'zones' in yaml_node:
+            transform = self.ref_level.transform
+            for zone_name, zone_yaml in yaml_node['zones'].items():
+                transform = self.levels[str(zone_yaml['level'])].transform
+                self.zones[zone_name] = \
+                    Zone(zone_yaml, zone_name, transform, self.coordinate_system)
 
     def parse_geojson(self, json_node):
         self.levels = {}
@@ -370,6 +380,7 @@ class Building:
             g['levels'] = {}
             g['lifts'] = {}
             g['doors'] = {}
+            g['zones'] = {}
 
             if self.coordinate_system == CoordinateSystem.web_mercator:
                 g['crs_name'] = self.global_transform.crs_name
@@ -390,6 +401,7 @@ class Building:
                 g['levels'][level_name] = level_graph
                 if level_graph['lanes']:
                     empty = False
+                    self.generate_zone_vertices_and_lanes(level_name,level_graph)
 
                 for door_edge in level.doors:
                     door_edge.calc_statistics(level.transformed_vertices)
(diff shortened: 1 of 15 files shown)


mxgrey:
I'm beginning the review of Stage A by looking through the zone feature in the editor. I have a few initial thoughts about the user experience:

* I find it's very easy to run into situations where the UI makes it impossible to delete one of the entry/exit lanes. For example if you delete a vertex in the zone and then try to delete its entry/exit lanes:

image

---

* When you click to edit an existing zone, the `Delete...` button of the first vertex will be focused. If the user pressed the Enter key while editing any of the values, they will accidentally delete the first vertex from the zone. It would better if the `OK` button is focused instead so that when the user presses enter, the zone editing tool registers the changes and closes:

image

* I also found it confusing that if I start to fill in a row with information and then click `Add...` afterwards, the information that I put in gets wiped out. From a design language standpoint, this tells the wrong thing to the user, since most users will interact with the table from left to right. Consider removing the `Add vertex...` row and providing a different way to add and delete the zone vertices.

---

* It took me a while to figure out the relationship between the two different tables in the `Zone Properties` dialog. I also found it confusing that I couldn't check both `is exit lane` and `is entry lane` for the same (`Internal vertex`, `External vertex`) pair. Please consider the following ideas:
  * Is there a fundamental reason that an internal vertex can't use the same external vertex for both entry and exit? Note that the dialog allows one external vertex to be used for entry of one internal vertex and exit of another internal vertex. Is that meant to be allowed, or is it just not being caught currently? Let's be as general as possible about what the tool allows. If there are bad practices that we should try to prevent then those should be warning pop-up dialogs instead of completely blocking the edit.
  * Consider combining the two tables into one, like:

| Vertex name | X (m) | Y (m) | Group | Priority | Entry vertex | Exit vertex |
|-|-|-|-|-|-|-|
| | | | | | | |

jixiancgh:
Hi @mxgrey , I've fixed some of the issues that you mentioned:
- Deleting a vertex will not block the deletion of their transition lanes
- The `OK` button is now focused instead of the `Delete...` button of the vertices table
- The `Add...` button row is now blocked to prevent users from adding information before clicking on the button 
- The `Add...` button is now arranged at the leftmost column

> Note that the dialog allows one external vertex to be used for entry of one internal vertex and exit of another internal vertex. Is that meant to be allowed, or is it just not being caught currently?

- Previously when there were two or more internal vertices, the validation logic finds a valid pair of transition lanes from vertex B and returns true even if vertex A has both entry and exit pointing to the same external vertex. This resulted in inconsistent validation of the transition lanes, such as the ones you have mentioned. This has been fixed.

---

> Is there a fundamental reason that an internal vertex can't use the same external vertex for both entry and exit? 

To answer this, I will use an example scenario where `deliveryBot_1` is trying to enter zone `test` from the external vertex `test_entry` and `tinyBot_1` is trying to exit from the zone to `test_exit`. In the current design, a robot inside a zone and a robot waiting at a zone entry point (after successfully obtaining a zone booking) are both set to stubborn. If there is only one transition lane for both entry and exit, the two robots will enter a deadlock and neither will yield, as they are blocking each other's only path.
Screenshot from 2026-04-21 11-09-37

To prevent the above scenario, users are required to choose **at least two external vertices** for each internal vertex, and have **at least one entry lane and exit lane from two different external vertices**. This should typically avoid the above situation by allowing the robot inside a zone to move to a "zone exit point" that's different from the "zone entry point".

image
			
---

> * Consider combining the two tables into one, like:

In the current design, we allow multiple entry and exit lanes. I've considered combining the two tables, but it resulted in too much clutter and probably not intuitive for users:
image

Instead, I've made some adjustments to the current design, shifting the `Show vertices` checkbox to the right, to hint that table is for vertices. Other than that, I have also added a warning when users tries to add a lane when there are no internal vertex available in the zone.
image

> The warning that pops up
image

What are your thoughts on this?

mxgrey:
Thanks for all your patience while we find time to review these PRs.

I think the current state of this PR is generally good. I just have a few more small pieces of feedback. Item (1) is a strong recommendation, while (2) and (3) are more for consideration and discussion.

### 1. Cancel is automatically focused when editing zone properties

Whenever I select the line-edit of a zone property, the cancel button comes into focus, like this:

image

Because of this, if I hit `enter` while editing anything, the zone editing window will be closed and any edit I did gets discarded.

I see that [this line](https://github.com/open-rmf/rmf_traffic_editor/pull/544/changes#diff-76649ce312927a5919d9a993b4f0a0410bdab779f1c7dfb805e53db46df30791R309) immediately sets the focus to the OK button, but that only matters when the widget is first spawned. To make `OK` the button that gets triggered whenever a user presses enter, you also need to add

```
  _ok_button->setDefault(true);
```

### 2. Consider giving a different color to the zone internal vertices

Since the internal vertices of zones cannot be used like regular vertices, users may get confused if they try to connect lanes to them. They'll actually be creating a new vertex inside of the zone, but the new vertex is not considered an internal vertex of the zone. That may even be doubly confusing.

I like that the zone entry/exit lanes are a different color from normal lanes. Perhaps the same gray color could be used for internal vertices?

### 3. Consider warning / preventing zone internal vertices from being outside the zone

Currently there is nothing to stop a user from placing an internal vertex of a zone outside of the dimensions of the zone, e.g.:

image

I think this is worth triggering a warning, both while editing and while generating the nav graph.

Similarly we might want to emit a warning when generating a nav graph that has a regular vertex on the inside of a zone.

jixiancgh:
Hi Grey, thanks for your suggestions, I have implemented them and some other fixes as well.

### 1. "OK" is now focused on default when editing zones
image

### 2. Internal vertices color has changed to differentiate them from regular vertices
image

### 3. Warnings in traffic editor
image

**^ When internal vertices are out of bounds**

image

**^ When transition lanes configuration might cause issues (only one connecting entry/exit lane to external vertex, no connecting lanes, etc)**

I have also removed the additional checks happening when deleting a transition lane. In general, the zone dialog will only warn the users if the configuration might have issues, and not reject the users' intended changes.

### 4. Warnings when building map
image

When building a zone with a regular vertex inside the zone
```
WARNING: regular vertex [(unnamed)] at index [12] in nav graph [0] lies inside zone [consultation]
```

When building a zone with the internal vertex outside of the zone
```
WARNING: zone [consultation] internal vertex [consultation#C#p3#wait_3] lies outside the zone bounds
```

These two warnings will be generated when building a map, and visible in `stdout.log`.

Other than that, I have caught some bugs and fixed them, like zone elevations not initialized for new zones, and zone table not resizing after newly adding a zone.

mxgrey (commented):
Thanks for all the conversations and refinement. I think this is good, I just need you to merge [these changes](https://github.com/chart-sg/rmf_traffic_editor_main/pull/5) to help the CI pass, then we can merge this.

jixiancgh:
Hi Grey, I've merged the changes and the CI should be good to go. Thanks for reviewing!

PR #545, opened 2026-06-12
Enabled Delete property button to the custom parameters in vertices GUI

DoppiaEffe94:
### Implemented feature

In the traffic-editor GUI, a user can add a property, just by clicking on the `Add property...` button.
However, on the other side, the `Delete property` does not work if the user wants to delete a custom parameter. This button seems disabled, with just a TODO in the [gui/editor.cpp](https://github.com/open-rmf/rmf_traffic_editor/blob/main/rmf_traffic_editor/gui/editor.cpp#L1356C1-L1362C2) and a message to invite the user to modify by hand the YAML.
This feature solves that TODO code section in a few ways:

- add `setSelectionBehavior(QAbstractItemView::SelectRows)` in order to highlights the whole row and make it clear to the user which params to delete.
 
- `connect(property_editor, &QTableWidget::itemSelectionChanged,[this](){...})` whenever the user select a new item, the connect() checks if the column 0 of the current row has `Qt::UserRole == true`. If yes, the button is enabled; otherwise, it stays disabled.
 
- only the custom params rows get the flag `Qt::UserRole == true`. This allows to delete only the custom parameters, and not touching the default ones.

- only the specific property in the .building.yaml file will be deleted.

Below a picture of a **no removable param** in GUI:

traffic_editor_not_removable_param

And here a picture of a **removable param** in GUI:

traffic_editor_removable_param

### GenAI Use
We follow [OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

- [x] I did not use GenAI

### DIFF
--- a/rmf_traffic_editor/gui/editor.cpp
+++ b/rmf_traffic_editor/gui/editor.cpp
@@ -232,9 +232,24 @@ Editor::Editor()
   property_editor->verticalHeader()->setSectionResizeMode(
     QHeaderView::ResizeToContents);
   property_editor->setAutoFillBackground(true);
+  property_editor->setSelectionBehavior(QAbstractItemView::SelectRows);
   connect(
     property_editor, &QTableWidget::cellChanged,
     this, &Editor::property_editor_cell_changed);
+  connect(
+    property_editor, &QTableWidget::itemSelectionChanged,
+    [this]()
+    {
+      const int row = property_editor->currentRow();
+      if (row < 0)
+      {
+        delete_param_button->setEnabled(false);
+        return;
+      }
+      QTableWidgetItem* item = property_editor->item(row, 0);
+      delete_param_button->setEnabled(
+        item && item->data(Qt::UserRole).toBool());
+    });
 
   QHBoxLayout* param_button_layout = new QHBoxLayout;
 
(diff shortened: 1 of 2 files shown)


luca-della-vedova (commented):
Thank you (o grazie?) for your contribution!

I gave it a try and it generally works but I found a few minor rough edges, I was wondering if you can reproduce / you think you can take a look at them (if not it's also OK, this is still better than the status quo).

Video below

https://github.com/user-attachments/assets/d0668c20-5643-4363-9615-a2d3b5e5b0a5

At 5 seconds, after deleting a property the `Delete property` button is still highlighted, clicking it again deletes the last property in the list.
I think a better behavior could be to either deselect the property (so the button becomes greyed out), or make sure the automatically selected property is highlighted.

As I delete more properties the level canvas itself is not updated until I finally click on the map itself and all the property icons are removed at once.
Is it possible to update it every time a property is deleted, or every time a property that has an icon is deleted?

DoppiaEffe94:
Ciao Luca,

I checked the rough edges that you were mentioning me and now they are solved.

Furthermore, I saw also another edge case, where, after selecting the custom parameter and then clicking in the empty space, both it does not disable the Delete property button and keeps the parameter highlighted in blue.
I fixed also this minor issue and now the situation is this as below:

https://github.com/user-attachments/assets/0bb48d5d-51ee-4794-921b-a653b02da6ad

Please, tell me if you find other edge cases and in case I will fix them.

luca-della-vedova (dismissed):
Very nice, thank you! Gave it a try and it works great :)

rmf_traffic_editor/gui/editor.cpp
```diff
@@ -681,6 +697,24 @@ void Editor::restore_previous_viewport()
   }
 }
 
+bool Editor::eventFilter(QObject* obj, QEvent* event)
+{
+  if (obj == property_editor->viewport() &&
+      event->type() == QEvent::MouseButtonPress)
```
mxgrey: I agree that this is how it should be formatted, but unfortunately [uncrustify is complaining about it](https://github.com/open-rmf/rmf_traffic_editor/actions/runs/27836309921/job/82706404563?pr=545#step:10:7429), causing all the CI to fail.

Let's just remove this indentation to unblock the CI, even though the formatting will look strange 🤷

mxgrey (changes requested):
Thanks for this long-needed contribution! There's just one small code formatting change that needs to happen for CI to pass.

DoppiaEffe94:
should be fixed now :)

mxgrey (approved):
The tsan failures were already present before this PR, so we'll ignore that.

Thanks for the contribution!

